PR #3, opened 2015-05-07
Support indexable selectors

juanazam:
Implement index option on attributes and queries to make definitions more dev friendly.

### DIFF
--- a/tests/unit/queries/attribute-test.js
+++ b/tests/unit/queries/attribute-test.js
@@ -55,3 +55,11 @@ it('uses page scope', function(assert) {
 
   assert.equal(attr(), 'Logo small');
 });
+
+it('searches for element by index if provided', function(assert) {
+  fixture('<img alt="img1" class="img"/><img alt="img2" class="img"/>');
+
+  var attr = buildAttribute(attributeAttribute, 'alt', '.img', { index: 2 });
+
+  assert.equal(attr(), 'img2');
+});
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -69,6 +69,18 @@ var page = PO.build({
 The following is a comprehensive documentation of the available `PO` attribute
 helpers.
 
+A set of options can be passed as parameters while defining attributes.
+
+```js
+var page = PO.build({
+  title: PO.text('.title', { index: 2 })
+});
+```
+
+The `index` option can be used when trying to specify a particular element,
+when the selector matches more than one element.
+In this example, the attribute will reference the second title of the page.
+
 ## Predicates
 
 Test conditions on elements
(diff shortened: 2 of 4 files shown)


san650:
This is great!

I think we can support the `index` option in all attributes, not only on the query ones. Would you mind adding the support for the rest?

If we resolve the selector inside `qualifySelector` helper then the option is available for all helpers.

tests/unit/queries/attribute-test.js
```diff
@@ -55,3 +55,11 @@ it('uses page scope', function(assert) {
 
   assert.equal(attr(), 'Logo small');
 });
+
+it('searches for element by index if provided', function(assert) {
+  fixture('<img alt="img1" class="img"></img><img alt="img2" class="img"></img>');
```
san650: `` tags doesn't support child elements. Would you mind changing the fixture tu use `<span>` elements instead?

test-support/page-object/queries.js
```diff
@@ -2,11 +2,23 @@
 
 import { qualifySelector, trim } from './helpers';
 
+function indexedSelector(selector, index) {
+  var iSelector;
```
san650: `iSelector` :crying_cat_face:

Maybe we could change `selector` to `baseSelector` and `iSelector` to `selector`?

juanazam: :+1: I wasn't inspired while thinking of that variable name :)

tests/unit/queries/attribute-test.js
juanazam: I'm not sure if I follow, you don't have to use nested child elements to use the index options, its just a way to select from any matched elements by a certain selector.

juanazam:
Thanks for your reply!

I didn't do it there because of the signature of that method (https://github.com/san650/ember-cli-page-object/blob/master/test-support/page-object/helpers.js#L3).

I wasn't sure if I should change the signature to something like:
```js
function qualifySelector(selector, ...scopes)
```
or assume that the last element of the parameters on the current signature is the one the index rule should apply (I think this is more error prone), which do you think is the best way?

Cheers!

tests/unit/queries/attribute-test.js
san650: What I meant was that "</img>" is invalid. It should be "".

san650:
Maybe `function qualifySelector(selector, scope, options = {})`?

juanazam:
Doing that would reduce how much the function can do:
right now it supports calls like: `qualifySelector('.my-outer-div', '.my-inner-div', '.my-element')`

In order to accomplish something similar with that signature: `qualifySelector('.my-outer-div .my-inner-div', '.my-element')`

With literals is not as troublesome as with variables, in that case, the call would look something like this:
`qualifySelector("${outerDiv}  ${innerDiv}", '.my-element')`

If we are ok with this, I'll implement it this way.

san650:
+2

san650:
Please, rebase. I've refactored a bit the code, so you may need to update your changes.

tests/unit/queries/attribute-test.js
juanazam: :+1:

juanazam:
Rebased and attended comments :smile:

san650:
:heart:

juanazam:
:smile: :smile: :smile: :smile: :smile: :smile:

PR #14, opened 2015-05-19
Define attributes without a selector if the page has an scope defined

### LINKED ISSUE #7: Define attributes without a selector if the page has an scope defined
Sometimes is handy to create a page or component and be able to define child attributes without setting a selector. For example:

```html
<div class="modal">
  <!-- modal content -->
</div>
```

```js
let page = PageObject.build({
  modal: component({
    scope: '.modal',

    isVisible: isVisible(),
    isHidden: isHidden(),
    isError: hasClass('error')
  })
});
```

In this case, the attributes `isVisible`, `isHidden` and `isError` validate properties on the `.modal` element. 

__Right now this works but is not officially supported, so we need to create tests and documentations for this behavior.__

As an added bonus, this definition can be improved using the new ES6 object literal shorthand assignments:

```js
let page = PageObject.build({
  modal: component({
    scope: '.modal',

    isVisible,
    isHidden,
    isError: hasClass('error')
  })
});
```

juanazam:
Closes #7

### DIFF
--- a/tests/unit/components/component-test.js
+++ b/tests/unit/components/component-test.js
@@ -8,6 +8,10 @@ import {
 } from '../test-helper';
 import { componentAttribute } from '../../page-object/build';
 import { textAttribute } from '../../page-object/queries';
+import {
+  isHiddenAttribute,
+  isVisibleAttribute
+} from '../../page-object/predicates';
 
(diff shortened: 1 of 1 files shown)


san650:
Thanks!

PR #16, opened 2015-05-26
Add selectable helper

### LINKED ISSUE #13: PO.selectable
For setting a value on a select field..

juanazam:
Closes #13

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -375,6 +375,37 @@ andThen(function() {
 });
 ```
 
+### `.selectable`
+
+Selects an option.
+
+Attribute signature
+
+```js
+PO.selectable(selector [, scope: ''])
+```
+
+Examples
+
+```html
+<select id="gender">
+  <option>Male</options>
+  <option>Female</options>
+</select>
+```
+
+```js
+var page = PO.build({
+  selectGender: PO.selectable('#gender')
+});
+
+page.selectGender('Female');
+
+andThen(function() {
+  // the option is selected
+});
+```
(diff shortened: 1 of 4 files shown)


DOCUMENTATION.md
```diff
@@ -375,6 +375,37 @@ andThen(function() {
...
+Attribute signature
+
+```js
+PO.selectable(selector [, scope: ''])
+```
+
+Examples
+
+```html
+<select id="gender" />
```
san650: Don't close the select tag

tests/acceptance/user-list-test.js
```diff
@@ -23,20 +23,25 @@ var page = PO.build({
     itemScope: 'tbody tr',
     item: {
       userName: PO.text('td', { index: 1 }),
-      role: PO.text('td', { index: 2})
+      role: PO.text('td', { index: 2}),
+      selectGender: PO.selectable('select'),
+      gender: PO.text('select option:selected')
```
san650: I think the right way of getting the selected options is using `PO.value`

We're using `jQuery.val()` under the hood so:
https://learn.jquery.com/using-jquery-core/faq/how-do-i-get-the-text-value-of-a-selected-option/

juanazam: Using `PO.value` would return the value attribute from the option, and we want just the text since options are most likely used without the value attribute when using ember.

san650: :+1:

juanazam:
:smile:

PR #20, opened 2015-06-01
Allow Components to act like a promise so async/await works

drewchandler:
The way actions are chainable is awesome, but one downside is that you can't use ES7's `async`/`await` with page objects, since actions return `this` instead of a promise.

This change makes it so that `Component` can act like a promise by implementing a `then` method. This allows you to do the following:

```javascript
test('login', async function(assert) {
  assert.expect(2);

  await page
    .visit()
    .form()
    .userName('user@example.com')
    .password('secret')
    .click();

  assert.ok(page.notHasError(), 'Page doesn\'t have error');
  assert.equal(page.message(), 'Valid user!');
});
```

I wasn't sure of the best way to test this. I could write an acceptance test using async/await, but I am not sure what side-effects turning on Babel's `es7.asyncFunctions` feature would have for apps using this add-on.

Do you think this feature is worthwhile? And if so, how would you like to see this tested?

### DIFF
--- a/test-support/page-object/actions.js
+++ b/test-support/page-object/actions.js
@@ -27,19 +27,19 @@ function visitable(params = {}) {
     path = fillInDynamicSegments(path, params);
   }
 
-  this.page.lastPromise = visit(path);
+  visit(path);
 
   return this.page;
 }
 
 function clickable() {
-  this.page.lastPromise = click(this.qualifiedSelector());
+  click(this.qualifiedSelector());
 
   return this.page;
 }
 
 function fillable(text) {
-  this.page.lastPromise = fillIn(this.qualifiedSelector(), text);
+  fillIn(this.qualifiedSelector(), text);
 
   return this.page;
 }
@@ -50,7 +50,7 @@ function clickOnText(text) {
   // want to __always__ click on the __last__ element that contains the text.
   let selector = this.qualifiedSelector(`:contains("${text}"):last`);
 
-  this.page.lastPromise = click(selector);
+  click(selector);
 
   return this.page;
 }
--- a/test-support/page-object/build.js
+++ b/test-support/page-object/build.js
@@ -1,7 +1,13 @@
+/* global wait */
+
 import { isNullOrUndefined } from './helpers';
 
 function Component() {
 }
+
+Component.prototype.then = function() {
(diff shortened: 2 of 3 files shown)


san650:
I think supporting `async/await` is a must, thanks for taking the time to implement this.

I like the implementation it's nice and simple. My original intention was to make the components behave like a promise by proxying the last promise's `then`, `catch` and `finally` methods.

You can see how __actions__ attributes are saving the last promise.
i.e.
https://github.com/san650/ember-cli-page-object/blob/96939a3ca35ebd4f59345e3e0e62d630e45c6010/test-support/page-object/actions.js#L30

But, what you did it's even better, I forgot that we can use the `wait` helper and let ember testing do the synchronization for us :smile: 

About testing, I think we could add a unit test for each action to test that they respond to `.then` because the only attributes we want to behave as promises are those.

About Babel's `es7.asyncFunctions` it would be great to have an acceptance test that uses the async/await keywords, but as you said, we don't know how it would affect apps that don't use that configuration. I'm going to try this on an existing app I have and let you know. This is not a must so we can add the async/await acceptance test later on a following up PR.

If you want, you can clean up the `page.lastPromise = ...` we're doing on the actions attributes, we're not going to need them :wink:.

drewchandler:
Unfortunately, `buildAttribute` doesn't use a real page object, so I can't test the `then` behavior in the action unit tests. For now, I just added an acceptance test where I call then on a chain of actions.

san650:
Thanks for adding this feature! It's going to be included in the next release.

san650:
This is included now on v0.5.0. Sorry the super late release!

brancusi:
Hi, can anyone explain how to use this properly? When attempting:

```js
await page.visit('posts');

//assert some element on screen
```
The `page.visit` does not have a then function on it and the assert fails. Did I misunderstand this PR?

After looking at the current tests, seems like this test was removed?

Any guidance on using async await would be great.

Thanks

san650:
@brancusi this is a regression introduced in the 1.0 release. Can you open a bug report?

san650:
@brancusi I've released [v1.0.1](https://github.com/san650/ember-cli-page-object/releases/tag/v1.0.1) which fixes async/await support.

Issue #22, opened 2015-06-06
Custom helpers


san650:
One of the problems we have right now with custom components is that we don't have information about the surrounding context, making it difficult to have app specific but generic components.

i.e.

```js
var input = function(selector) {
  return component({
    value: value(selector),
    hasError: hasClass('is-error', selector)
  });
}

var page = PageObject.build({
  search: input('.header input'),

  users: collection({
    itemScope: 'tr',
    item: {
      // This won't work because `input` helper
      // doesn't have context awareness, so the
      // the selector don't get prefixed with the current row.
      userName: input('input:first')
    }
  })
});

// If we take a look at the __private__ `qualifedSelector`
page.users(3).userName().qualifiedSelector() // => `input:first` and not `tr:nth-of-type(3) input:first`
```

We need to have a way to define custom helpers that can have context awareness

Something like:

```js
var input = customHelper(function(selector, scope, options) {
    return {
        scope: scope,
        value: value(selector, options)
    };
});

var page = PageObject.build({
  search: input('.header input'),

  users: collection({
    itemScope: 'tr',
    item: {
      // This would work fine now
      userName: input('input:first')
    }
  })
});

// If we take a look at the __private__ `qualifedSelector`
page.users(3).userName().qualifiedSelector() // => `tr:nth-of-type(3) input:first`
```

PR #26, opened 2015-06-07
Implement custom helpers

juanazam:
Implements #22

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -621,7 +621,11 @@ Note that if the plain object doesn't have attributes defined, the object is ret
 Defines a way in which custom components can be created using
 information of their surrounding context.
 
-It expects a function with the following signature:
+```js
+PO.customHelper(userFunction);
+```
+
+Where the provided user function has the following signature:
 
 ```js
 function(selector, options) {
@@ -668,21 +672,21 @@ Here is another two examples of using `customHelper` but this time, the
 user defined functions return a function and a string.
 
 ```js
-var admin = PO.customHelper(function(selector, options) {
+var isAdminField = PO.customHelper(function(selector, options) {
   return function() {
     return $(selector).hasClass('admin');
   };
 });
 
-var disabled = PO.customHelper(function(selector, options) {
+var isDisabled = PO.customHelper(function(selector, options) {
   return $(selector).prop('disabled');
 });
 
 var page = PO.build({
   myText: {
     scope: '#myInput',
-    isAdminField: admin(),
-    isDisabled: disabled()
+    isAdminField: isAdminField(),
+    isDisabled: isDisabled()
   }
 })
 ```
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -615,6 +615,44 @@ var page = PO.build({
 
 Note that if the plain object doesn't have attributes defined, the object is returned as is.
 
+### `.customHelper`
+
+Defines a way in which custom components can be created using
+information of their surrounding context.
+
+It expects a function with selector, scope and options
+as parameters which the user can use as context when creating
+the custom component.
+
+Here is an example of usage:
+
+```js
+var input = customHelper(function(selector, scope, options) {
+    return {
+        scope: scope,
+        value: value(selector, options)
+    };
+});
+
+var page = PageObject.build({
+  search: input('.header input'),
+
+  users: collection({
+    itemScope: 'tr',
+    item: {
+      // This would work fine now
+      userName: input('input:first')
+    }
+  })
+});
+```
+
+If we execute:
+
+`page.users(3).userName().qualifiedSelector()`
+
+It would return `tr:nth-of-type(3) input:first`
+
 ## Attribute options
 
 A set of options can be passed as parameters when defining attributes.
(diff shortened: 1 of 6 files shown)


juanazam:
@san650 I have finished with the discussed suggestions, please check it out :D

test-support/page-object/custom-helper.js
```diff
@@ -0,0 +1,30 @@
...
+          let scopedSelector = qualifySelector(parent.scope, selector);
+
+          let customHelperRes = userDefinedFunction(scopedSelector, options);
+
+          if ($.isPlainObject(customHelperRes)) {
+            return build($.extend(customHelperRes, { scope: scopedSelector }),
+                         key,
+                         parent);
+          } else if ($.isFunction(customHelperRes)) {
+            return customHelperRes(scopedSelector);
```
san650: I think we don't need to pass the scope as an argument here. We just need to invoke the function, right?

test-support/page-object/custom-helper.js
```diff
@@ -0,0 +1,30 @@
...
+        return function() {
+          // eval
+          let scopedSelector = qualifySelector(parent.scope, selector);
+
+          let customHelperRes = userDefinedFunction(scopedSelector, options);
+
+          if ($.isPlainObject(customHelperRes)) {
+            return build($.extend(customHelperRes, { scope: scopedSelector }),
+                         key,
+                         parent);
```
san650: I think the order of the parameters in the `extend` is not the best.

If the customHelperRes has a `scope: 'something'` attribute, we're overriding it with our `{ scope: scopedSelector }` and may not be the desired effect.

I think the order could be `$.extend({ scope: scopedSelector }, customHelperRes)`, this way the user can override the scope.

test-support/page-object/custom-helper.js
```diff
@@ -0,0 +1,30 @@
...
+  return function(selector, options) {
+    // def
+    return {
+      buildPageObjectAttribute: function(key, parent) {
+        // def
+        return function() {
+          // eval
+          let scopedSelector = qualifySelector(parent.scope, selector);
+
+          let customHelperRes = userDefinedFunction(scopedSelector, options);
```
san650: Maybe we can simplify the code by renaming some variables `s/customHelperRes/response` `s/scopedSelector/qualifiedSelector/`

Also, maybe we can improve readability by having only one return statement

```js
if ($.isPlainObject(response)) {
    processedResponse = build(...);
} else if ($.isFunction(response)) {
    processedResponse = response();
} else {
    processedResponse = response;
}

return processedResponse;
```

What do you think?

DOCUMENTATION.md
```diff
@@ -615,6 +616,87 @@ var page = PO.build({
 
 Note that if the plain object doesn't have attributes defined, the object is returned as is.
 
+### `.customHelper`
+
+Defines a way in which custom components can be created using
+information of their surrounding context.
+
+It expects a function with the following signature:
```
san650: If you look at the other attributes, we add a _attribute signature_ section after the short description.

Something like...

```markdown
Attribute signature

` ``js
PO.customHelper(userFunction);
` ``

Where the provided user function has the following signature

` ``js
function(selector, options) {
  // user __code__ goes here
  return Value;
}
` ``

...

```

Or something on the line

DOCUMENTATION.md
```diff
@@ -615,6 +616,87 @@ var page = PO.build({
...
+
+var disabled = PO.customHelper(function(selector, options) {
+  return $(selector).prop('disabled');
+});
+
+var page = PO.build({
+  myText: {
+    scope: '#myInput',
+    isAdminField: admin(),
+    isDisabled: disabled()
```
san650: Can we change `disabled` and `admin` function names to be `isDisabled` and `isAdmin`? So we follow the same convention as the rest of the project

san650:
@juanazam This is great! Sorry I did a lot of comments but I think they're just small adjustments.

I'm going to test the functionality on a real project now and let you know how it goes :smile:

test-support/page-object/custom-helper.js
juanazam: Following the spec we agreed on, we must pass the result of calling `qualifySelector(parent.scope, selector)` to get the parent's scope when the user defined function is called.

```
var input = customHelper(function(selector, options) {
  let hasError = $(selector).parent().hasClass('has-error');

  return {
    value: value(selector),
    hasError: function() {
      return hasError;
    }
  };
});

var page = PageObject.build({
  scope: 'form',
  userName: input('#userName')
});
```
In this case, the input function should receive `form #userName` as a parameter.

test-support/page-object/custom-helper.js
juanazam: :+1:

test-support/page-object/custom-helper.js
juanazam: :+1:

DOCUMENTATION.md
juanazam: :+1:

juanazam:
@san650 I addressed all issues except the discussion around if the selector should be passed as a parameter when the user defined function returns a function. :smile:

PR #28, opened 2015-06-25
Normalize whitespaces in the way browser does it

irnc:
This change will allow to write page objects without knowledge of how text is broken down on multiple lines, making tests less fragile.


juanazam:
Hey @irnc thanks! expect this to be included on our next release :smile:

san650:
This is included now on v0.5.0.

Thanks for contributing!

PR #30, opened 2015-06-26
Add NPM package metadata

san650:


san650:
The idea with AUTHORS file is that NPM loads this file and fills the `contributors` field of the package on npmjs.org.

More info on https://docs.npmjs.com/files/package.json#people-fields-author-contributors

@juanazam please review

PR #40, opened 2015-07-15
Provide queryParam support for visitable action

donovan-graham:
Query params are optional and would just be tacked onto the end of the existing url, if they are specified.


san650:
Thanks for contributing!

san650:
Released to NPM in v0.6.0 https://github.com/san650/ember-cli-page-object/releases/tag/0.6.0

PR #42, opened 2015-07-15
Improve collection item selector

### LINKED ISSUE #41: Improve collection item selector
Given the following HTML

```html
<ul>
  <li class="original">Original</li>
  <li>Item 1</li>
  <li>Item 2</li>
</ul>
```
In order to select all items except the original one I would build a PO collection which looks like:

```javascript
items: collection({
  scope: 'li:not(.original)',

  item: {
  }
})
```  

In order to access the items, I would have to write something like ```page.items(1)```, but this is not working with the current implementation because the selector will be ```$("li:not(.original):nth-of-type(1)")``` and the result will be empty in this case.

acostami:
Fixes #41

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -905,7 +905,7 @@ var page = PageObject.build({
 |  | translates to |
 | ------ | -------- |
 | `page.todos().create()` | `click('.todo button')` |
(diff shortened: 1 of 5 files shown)


juanazam:
:+1: :+1: :+1: :+1:

san650:
Thanks for contributing!

PR #46, opened 2015-08-12
fix example for isHidden

csantero:


juanazam:
hey @csantero thanks for this, I'll merge it right away

Issue #56, opened 2015-09-14
Assert selector matches single element?


machty:
I trolled myself earlier because I chose a `.text` selector that matched more than one element (awkwardly enough it matched the desired element, as well as a parent/ancestor element of that desired element), and when I asserted on the value, it kept telling me the value was "Login  Login". I thought it was a cli-page-object bug until I dug in more.

My foolish class-naming aside, it seems very unlikely that you'd want page object components to match multiple elements, so perhaps this library should assert that exactly one element is matched?

san650:
Hi @machty, thanks for your feedback.

I'm not sure if we should assert that exactly one element is matched or if we need to use only one of the matched elements. For some actions (i.e. `clickOnText`) we use the last matched element of the matched set.

So, I'm not sure if this is would be the desired behavior for _all_ the attributes, maybe we should make this the default behavior for predicates and query attributes but not for actions.

So, `.hasClass`, `.notHasClass`, `.isVisible`, `.isHidden`, `.attribute`, `.count`, `.text` and `.value` attributes could raise an error when more than one element is matched.

What do you think?

machty:
Is it desirable for `clickOnText` to use the last matched element? That seems a bit like a footgun. If you're performing an action, isn't the expectation that you'd only be performing it on one element?

juanazam:
@san650 I would leave `.count` out of that list, we are using it to count how many elements of the selector passed as parameter exists so I guess it doesn't make a lot of sense to assert in that case.

@machty Regarding `clickOnText`, using the last matched element is an arbitrary decision, the thing is that there might me multiple elements with the same text to be clicked, several "Login" button for instance. I believe `clickOnText` should be used when there is no difference between clicking any element with the "Login" text. If I want to click a particular element, I would use `.click` and provide a unique selector, in that case it might be a good idea to assert it only matches one element.

What do you think?

san650:
I think `clickOnText` is a special case. We use the last matched element because we are looking for elements that contains a specific text. Let me try to explain it with an example.

Suppose that you have a modal with two buttons

```html
<div class="modal">
  <div class="modal-body">
    <p>Lorem ipsum</p>

    <button>Accept</button> <button>Cancel</button>
  </div>
</div>
```

You could have a page object like

```js
var modal = PageObject.build({
  click: clickOnText('.modal');
});

modal.click('Accept');
```

The selector we're generating for the `modal.click` call is `.modal
[...]

PR #65, opened 2015-10-13
First attempt at asserting single element matchings

juanazam:
This PR attempts to show some code of how #56 could be implemented trying to maintain backwards compatibility with the old behavior by using an override options.

I would like to hear @san650 and @machty input on this.

Cheers!

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -3,6 +3,7 @@
 Table of contents
 
 * [Setup](#setup)
+  * [Global config](#global-config)
 * [Predicates](#predicates)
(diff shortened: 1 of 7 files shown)


san650:
@juanazam I think it would be easier to target this PR to milestone [v1.0](https://github.com/san650/ember-cli-page-object/milestones/v1.0) which won't require backward compatibility support.

machty:
Sorry for not piping up sooner but this looks good to me!

san650:
Implemented in #91

machty:
nice :)

PR #68, opened 2015-10-16
Nested collections are not properly scoped

backspace:
Hey, here is a test showing that nested collections aren’t inheriting the scope of their containing collection. I see that you [recently fixed something related to collection scope](https://github.com/san650/ember-cli-page-object/releases/tag/v0.8.1) and the changes look pretty in-depth, so I suspect you’d be much quicker at fixing this than me.

Or maybe I’m just mistaken and this is expected behaviour?


juanazam:
@backspace thanks for this! we will check what is going on as soon we can.

Cheers!

PR #72, opened 2015-10-19
Nested collection items do not use the correct parent index

backspace:
Hey, the recent changes helped with my use of `count()` on a nested collection, but when using an index to get an item in a nested collection, it’s always scoped to the first item of the parent collection.

Thanks so much for your work on this, I really like how it’s improving my tests.

### DIFF
--- a/tests/unit/components/collection-test.js
+++ b/tests/unit/components/collection-test.js
@@ -260,6 +260,7 @@ test('assigns the correct scope to sub collection items', function(assert) {
   ).toFunction();
 
   assert.equal(attribute(1).spans(2).text(), 'Ipsum');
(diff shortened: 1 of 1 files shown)


san650:
Good catch! I think this is the same issue than #73. I'll take a look at it.

Thanks for the test and the report!

san650:
Thanks for your contribution. See [v0.8.3](https://www.npmjs.com/package/ember-cli-page-object)

Release notes: https://github.com/san650/ember-cli-page-object/releases/tag/v0.8.3
Npm: https://www.npmjs.com/package/ember-cli-page-object

backspace:
Excellent, fixed again, thanks!

Issue #63, opened 2015-10-08
Should component scope inside a collection preserve parent selector?


leonardocsouza:
Hi there!

I've just recently started using this awesome addon (thank you so much!) and noticed what seems to be an issue, but I wanted to confirm if it's indeed one.

I have a situation where I have a tree of nested elements and all the scoping for each collection of items within another collection works great, with exception of a case where I'm trying to use a component as an attribute of an item in a collection, but I can't find a way to find this component inside the item without manually appending an extra class on each property for the component. 

Consider this HTML below as the leaf node of my tree:
```
... lines omitted for brevity
<li class="topic">
    <div class="flagging-menu is-open">
        <div class="mark urgent">Urgent</div>
        <div class="mark current">Current</div>
        <div class="mark past">Past</div>
    </div>
</li>
... lines omitted for brevity
```

Then what I'm trying to do is to define a component for the `flagging-menu` within the`topic` item so that I can simply access the properties of the component without having to pass the component selector every time.

My first attempt was this:
```
topics:  collection({
  itemScope: '.topic',
  item: {
    title:      text('.title'),    

    flaggingMenu: {
      scope: '.flagging-menu',
      isOpen: hasClass('is-open'),

      selectCurrent: clickable('.mark.current'),
      selectPast: clickable('.mark.past'),
      selectUrgent: clickable('.mark.urgent')
    }
  }
})
```

The issue I ran into was that by setting `scope: '.flagging-menu'`, the selector for the component dropped the parent scope (`.topic`) and matched any item on the page matching the broader `.flagging-menu` selector, when what I would have liked it to do was to _append_ the component scope to the parent selector (i.e. `.topic .flagging-menu`).

The only way I found to get what I was trying to accomplish was to pretend I had a collection of `flagging-menu`, since by using `itemScope` I am able to preserve the parent scope while also narrowing down to the specific element I'm interested in.

```
topics:  collection({
  itemScope: '.topic',
  item: {
    title:      text('.title'),    

    // Hackish way to be able to have flaggingMenu preserve scope
    // of parent item. Using scope: attribute replaces parent scope.
    flaggingMenu: collection({
      itemScope: '.flagging-menu',

      item: { 
        isOpen: hasClass('is-open'),

        selectCurrent: clickable('.mark.current'),
        selectPast: clickable('.mark.past'),
        selectUrgent: clickable('.mark.urgent')
      }
    })
  }
})
```

[...]

PR #73, opened 2015-10-19
Assign correct scope to sub-collection items

HeroicEric:
Same issue as described in https://github.com/san650/ember-cli-page-object/issues/63

I believe that the test that was added was only passing because "Ipsum" was the value for `attribute(1).spans(2)` and also happened to be the same as `spans(2)` without the `attribute(1)` scope.

This correctly assigns the scope of `target` when an `index` is provided to `getCollection`.

### DIFF
--- a/test-support/page-object/collection.js
+++ b/test-support/page-object/collection.js
@@ -54,6 +54,10 @@ function getCollection(target, key, options, index) {
   }
 
   if (index) {
+    if (target.__forceScopeToChildren) {
+      options.scope = target.scope;
+    }
+
     component = copy(options.itemDefinition);
     component.scope = qualifySelector(options.scope, scopeWithIndex(options.itemScope, index));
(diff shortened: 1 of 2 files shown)


san650:
Thanks for the PR. I think this is the same issue @backspace found in #72.

I'll review it and test it thoroughly.

san650:
@HeroicEric this seems to fix the issue :)

I would like to do a bugfix release with this fix only. Would you mind to cherry-pick this commit to a new branch on top of [v0.8.x](https://github.com/san650/ember-cli-page-object/tree/v0.8.x) branch? There are a bunch of new stuff on master that I don't want to include on the bugfix release.

I can do it for you if you want, but I don't want to include a commit "on your behalf" unless you're ok with it.

Thanks!

san650:
Fixed by #74

PR #80, opened 2015-11-16
Add test showing that nested components lose scope

backspace:
Hello again! Still loving this.

I recently had some confusing test failures working with a collection of click-to-edit elements. The test was set to click the first element for editing but instead all elements were being clicked. It appears that doubly-nested components don’t inherit their parent collection scope.

Hopefully fixing this would fix it for even deeper nesting, maybe another test should be added for that?

### DIFF
--- a/tests/unit/components/collection-test.js
+++ b/tests/unit/components/collection-test.js
@@ -94,6 +94,28 @@ test('sets scope to components under item object', function(assert) {
   assert.equal(attribute(2).button().text(), 'Right');
 });
 
+test('sets scope to deeper components under item object', function(assert) {
+  fixture('<ul><li><main><header>Wrong Title</header></main></li><li><main><header>Right Title</header></main></li></ul>');
+
+  let attribute = buildProperty(
+    collection({
+      itemScope: 'li',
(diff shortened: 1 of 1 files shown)


juanazam:
@backspace thanks for this!, it's very useful, we will check into it.

juanazam:
@backspace I have investigated this issue, and it would be fixed by adding `attr.__forceScopeToChildren = true` after this line https://github.com/san650/ember-cli-page-object/blob/master/test-support/page-object/create.js#L70.

But we need to discuss if this is the behavior we want to support from now on (it seems to be the case). The thing is that in the past, setting the `scope` attribute on a component used to reset the component's scope (meaning that for example if the page had `scope` defined, using it in the component would override it).


@san650  We might need to think about which is the behavior we want to support regarding scope inheritance. In case of collections, it seems to make a lot of sense to inherit, and maybe for components too.

san650:
Just to follow up this issue: We're targeting the fix for the 1.0, you can see the progress on #91 and #92

san650:
Fixed by #91

PR #81, opened 2015-11-24
textList helper

kagemusha:
Allows you to get the text of a set of elements as an array.

E.g. given `<li>a</li><li>b</li>`
`textList('li')` => `['a', 'b']`


test-support/page-object/properties/text-array.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {Object} options - Additional options
+ * @param {string} options.selector - CSS selector of the element to check
+ * @param {string} options.scope - Overrides parent scope
+ * @return {array} array of text strings for the selected elements
+ */
+function getTextArray(target, key, options) {
+  let elements = findElementWithAssert(options, target);
+
+  let textStrings = new Array(elements.length);
+  for (var i=0, l=textStrings.length; i<l; i++) {
```
juanazam: What do you think about using `Ember.map` instead of iterating with `for`?

juanazam:
@kagemusha thanks for this, it's a good idea! I just left one comment, can you please take a look at it?

I'm a bit unsure about `textArray` as a name, what do you think about `allText` or `textList`?

test-support/page-object/properties/text-array.js
kagemusha: have tried this (as wasn't the original author of the function and also thought it strange), but `elements` is not a real array so `map`, etc. don't work

kagemusha:
yeah was unsure about `textArray` also.  of the other options i've considered, `textList` probably best.

juanazam:
Excellent, two more things:

1. We need to expose the `textList` property on the page object API, you can check how it's done here: https://github.com/san650/ember-cli-page-object/blob/master/test-support/page-object.js#L19.
2. Add yourself to the Authors file: https://github.com/san650/ember-cli-page-object/blob/master/AUTHORS.

Then we are done :smile:

kagemusha:
done and done.  thx to you guys for this project...been very useful

juanazam:
@kagemusha glad you like it and thanks for making it even better! I'll let you know when the new version with this change is released.

juanazam:
@kagemusha I have released v0.10.0 and this feature is included, you can update and use it :smile:, thanks for contributing.

kagemusha:
Cool.  Thx!

PR #82, opened 2015-11-28
Make components and collections inherit scope by default

juanazam:
Based on discussions started by #80, here is a first draft of the behavior described.

@san650 if we agree on this, we can start thinking on releasing v1.0 with this since it's a big behavior change.


backspace:
Thanks for the investigation and possible fix. I see how this could be a big change, though it seems like the expected behaviour to me.

If it’s too soon for 1.0, there’s always 0.10, 0.11, and beyond! :smile:

san650:
This is going to be implemented for the 1.0.

Closing this PR because this is going to be implemented in 1.0 branch.

backspace:
Thanks for the update, I look forward to 1.0!

Issue #54, opened 2015-09-02
Allow mapping of collections


richmolj:
Please correct me if I'm wrong, but I *think* it's not possible to do the following:

```javascript
page.users().mapBy('name')
```

Or elsewise get a list of `users` without supplying every index. This would be nice to verify the UI is showing the correct list of users.

san650:
Hi @richmolj, you're right, it's not possible to do that now. It's a nice idea and would be great to have this feature on collections.

san650:
Since a couple of releases we now have a new attribute `textList` which allows to retrieve the text of multiple items as an array.

http://ember-cli-page-object.js.org/api/queries/#textlist

I think this allows to do something similar to what this issue proposes.

```js
var page = PageObject.create({
  usersName: textList('tr td:first')
});

page.usersName() // => ['John', 'Jane']
```

We're changing the notation a bit for version [1.0](https://github.com/san650/ember-cli-page-object/milestones/v1.0), you can see an example here https://github.com/san650/ember-cli-page-object/blob/45b63b7c5a06e841cb83d301bbb2044465be8e8d/tests/unit/queries/text-test.js#L144-L158 although this is not merged yet.

```js
test('returns multiple values', function(assert) {
  fixture(`
    <ul>
      <li>lorem</li>
      <li> ipsum </li>
      <li>dolor</li>
    </ul>
  `);

  var page = PageObject.create({
    foo: text('li', { multiple: true })
  });

  assert.deepEqual(page.foo, ['lorem', 'ipsum', 'dolor']);
});
```

dfreeman:
@san650 The `{ multiple: true }` option is great for some cases, but I still often find myself wanting to check multiple features at a time of each item in a collection while testing.

Would you consider a PR that adds something like a `toArray()` convenience method on collections? That would allow people to iterate, map, and do whatever else to their heart's content without needing to drastically expand the existing API.

san650:
@dfreeman It would be great to start adding Enumerable methods to the collection. Right now we only have the `count()` method.

Although a `toArray()` solves all the problems, I would prefer to add `map(function(element) {}): Array` and `mapBy(propName): Array` functions. Right off the top of head I'd say that's what users need most of the time. Maybe we can include a `reduce` function also to tackle the rest of the cases.

What do you think?

san650:
Just to complement my prior comment, here's an example of how I imagine people could use the collections:

```js
[...]

PR #91, opened 2015-12-20
One point zero

san650:
This PR is work in progress!

# To do

## Required

- [x] Remove `build` method
- [x] Remove `component` property
- [x] Migrate code to use [Ceibo](https://github.com/san650/ceibo)
  - Allows us to use properties instead of functions  (**Breaks compatibility**)
  - Allows to navigate to parent properties
- [x] Make scopes inheritable by default  (**Breaks compatibility**)
- [x] Add `resetScope` option to all properties
- Remove custom helper  (**Breaks compatibility**)
  - [x] Remove code
  - [ ] Write examples of how to write similar helpers using descriptors
- [x] Make properties to throw an error when multiple elements are matched  (**Breaks compatibility**)
  - Except for `count` attribute :smile: 
- [x] Make collections zero-based (**Breaks compatibility**)
- [ ] Update website documentation
- [x] Rename property option `index` to `at`
- [x] Allow to require properties directly `import { create, text, count } from './page-object'`
- [x] isVisible fails when element is not present (#95)
- [ ] Add an easy way to extend existing page objects

## Nice to have

- [ ] Review `visitable` method signature (is there a better alternative?)
- [ ] Debug Mode (print in console everything similar to what `ENV.APP.LOG_RESOLVER` does)
- [ ] Support component integration tests
- [ ] Rethink how to interact with collection of values (#54, #81)
  - [x] Maybe we could support something like `text('span', { multiple: true })`


test-support/page-object/properties/collection.js
```diff
@@ -0,0 +1,75 @@
...
+import { create } from '../create';
+import { count } from './count';
+
+function merge(target, ...objects) {
+  objects.forEach(o => Ember.merge(target, o));
+
+  return target;
+}
+
+function select(target, ...keys) {
```
san650: Codacy Issue found: ["select" is defined but never used](https://www.codacy.com/app/san650/ember-cli-page-object/file/1924265012/issues/source?bid=2903707&fileBranchId=3041068#l12)

test-support/page-object/properties/value.js
```diff
@@ -1,42 +1,31 @@
-import Descriptor from '../descriptor';
 import { findElementWithAssert, trim } from '../helpers';
```
san650: Codacy Issue found: ["trim" is defined but never used](https://www.codacy.com/app/san650/ember-cli-page-object/file/1940277127/issues/source?bid=2903707&fileBranchId=3041068#l1)

PR #93, opened 2015-12-28
Allow nested pages with full path to "page-object" module

rtablada:
Currently making a page nested in a namespace leads to failing module look up because the blueprint uses `../page-object`.

So the command `ember g page-object parent/child` will fail silently if you try use this new page object generated.

This instead uses the same pattern as the Acceptance test generator from Ember CLI to use the fully qualified module name for the `page-object` module.

### DIFF
--- a/blueprints/page-object/files/tests/pages/__name__.js
+++ b/blueprints/page-object/files/tests/pages/__name__.js
@@ -1,4 +1,4 @@
-import PageObject from '../page-object';
+import PageObject from '<%= testFolderRoot %>/page-object';
 
 let {
   visitable
--- a/blueprints/page-object/index.js
+++ b/blueprints/page-object/index.js
@@ -1,3 +1,12 @@
+var stringUtils = require('ember-cli-string-utils');
+
 module.exports = {
-  description: 'Generates a page object for acceptance tests.'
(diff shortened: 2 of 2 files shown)


san650:
This is great! Thanks for contributing

blueprints/page-object/files/tests/pages/__name__.js
```diff
@@ -1,4 +1,4 @@
-import PageObject from '../page-object';
+import PageObject from '<%= testFolderRoot %>/page-object';
```
san650: I tested this locally and it seems that we're missing the `/tests` folder in the require statement. It should be something like:

`import PageObject from '<%= testFolderRoot %>/tests/page-object';`

Take a look at https://github.com/ember-cli/ember-cli/blob/2ccd785999f6f815c5738200128a16720bd3f4da/blueprints/acceptance-test/files/tests/acceptance/__name__-test.js#L2

san650:
@rtablada Apart from the issue I commented, this PR seems ready to go. Please address the issue and I'll merge it.

We also have an [AUTHORS](https://github.com/san650/ember-cli-page-object/blob/master/AUTHORS) file, can you update it?

Thanks!

san650:
Thanks for contributing!

PR #107, opened 2016-01-28
`.isVisible` does not raise an exception

### LINKED ISSUE #95: isVisible fails when element is not present
According to the documentation, `isVisible` should return `false` when the element is hidden OR the element does not exist. But the library currently throws an error when the element does not exist. I believe this is because the code ends up hitting `findWithAssert`: https://github.com/san650/ember-cli-page-object/blob/master/test-support/page-object/properties/is-visible.js#L16.

san650:
Fixes #95


san650:
wrong branch!

PR #109, opened 2016-01-31
Add mechanism to extend page objects

san650:
This PR adds an easy way to extend existing page objects

```js
var page = create({
  foo: 'bar'
});

var extended = page.extend({
  baz: 'qux'
});

assert.equal(extended.foo, 'bar');
assert.equal(extended.baz, 'qux');
```

### DIFF
--- a/test-support/page-object/create.js
+++ b/test-support/page-object/create.js
@@ -7,7 +7,7 @@ import { clickOnText } from './properties/click-on-text';
 import { clickable } from './properties/clickable';
 import { contains } from './properties/contains';
 
-var { merge } = Ember;
+var { merge, $ } = Ember;
 
 function plugDefaultProperties(definition) {
   if (typeof(definition.isVisible) === 'undefined') {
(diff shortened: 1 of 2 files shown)


san650:
@juanazam @jeradg I was thinking on something like this to support an extend mechanism, thoughts?

juanazam:
@san650 I think this is great! we can move it Ceibo later so we can use it of other use cases besides `ember-cli-page-object`.

jeradg:
@san650 @juanazam: I think this would work well.

One concern: my instinct is that the syntax should support the typical Ember `Object.extend({...}).create()` idiom. That way people don't have to learn a new syntax for creating objects.

I'm not sure if that would require a lot of extra work for no additional functionality, though. If it would be a pain to do, then this is probably fine.

san650:
We need to think a bit more the API. Closing this PR for now

sohara:
Any chance this will get back on the roadmap? It would be helpful to allow shared code in page objects.

san650:
@sohara I think the intention is to add a way to achieve this, what we're not sure is how. Can you open a feature request?

PR #110, opened 2016-01-31
Assert predicates and queries only match one element

juanazam:
* [x] Predicates and queries throw error when DOM lookups return more than one element.
* [x] `multiple: true` option overrides this behavior.

### DIFF
--- a/test-support/page-object/properties/count.js
+++ b/test-support/page-object/properties/count.js
@@ -22,6 +22,7 @@ export function count(selector, options = {}) {
     isDescriptor: true,
 
     get() {
+      options.multiple = true;
       return findElement(this, selector, options).length;
     }
   };
--- a/test-support/page-object/properties/count.js
+++ b/test-support/page-object/properties/count.js
@@ -1,4 +1,7 @@
+import Ember from 'ember';
 import { findElement } from '../helpers';
+
+var $ = Ember.$;
 
 /**
  * Gets the count of matched elements
@@ -22,8 +25,11 @@ export function count(selector, options = {}) {
     isDescriptor: true,
 
     get() {
-      options.multiple = true;
-      return findElement(this, selector, options).length;
+      let countOptions = {};
+
+      $.extend(true, countOptions, options, { multiple: true });
+
+      return findElement(this, selector, countOptions).length;
     }
   };
 }
--- a/test-support/page-object/properties/is-hidden.js
+++ b/test-support/page-object/properties/is-hidden.js
@@ -1,4 +1,7 @@
+import Ember from 'ember';
 import { findElement } from '../helpers';
+
+var $ = Ember.$;
 
 /**
  * Creates a predicate to validate if an element is hidden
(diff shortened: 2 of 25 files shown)


test-support/page-object/properties/is-hidden.js
```diff
@@ -22,9 +25,22 @@ export function isHidden(selector, options = {}) {
...
+        result;
+
+      if (element.length > 0) {
+        if (options.multiple) {
+          result = !(Ember.A(element).any(function(e) {
+            return $(e).is(':visible')
+          }));
+        } else {
+          result = element.is(':hidden')
+        }
```
san650: This can be written as `elements.get().reduce((result, elem) => result && $(elem).is(':hidden'));` or something like that without testing `options.multiple`. The `multiple` condition was already tested by `findElement`

test-support/page-object/properties/is-visible.js
```diff
@@ -22,9 +23,22 @@ export function isVisible(selector, options = {}) {
...
 
-      return element.is(':visible');
+      if (element.length > 0) {
+        if (options.multiple) {
+          result = !(Ember.A(element).any(function(e) {
+            return $(e).is(':hidden')
+          }));
+        } else {
+          result = element.is(':visible')
+        }
```
san650: Same case as `isHidden`

tests/unit/predicates/contains-test.js
```diff
@@ -65,11 +65,41 @@ test('resets scope', function(assert) {
...
+    <span>lorem</span>
+    <span>ipsum</span>
+    <span>dolor</span>
+  `);
+
+  let page = create({
+    foo: contains('span', { multiple: true })
   });
 
   assert.ok(page.foo('lorem'));
```
san650: I think `contains` should return `true` if every element contains that text (similar to how `isHidden` behaves) What do you think?

tests/unit/predicates/has-class-test.js
```diff
@@ -87,6 +87,36 @@ test('resets scope', function(assert) {
...
+    <span class="lorem"></span>
+    <span class="ipsum"></span>
+  `);
+
+  let page = create({
+    foo: hasClass('lorem', 'span', { multiple: true })
+  });
+
+  assert.ok(page.foo);
+});
```
san650: We need a test for the case when one element has the class and another element doesn't. It should return false in that case.

tests/unit/predicates/is-hidden-test.js
```diff
@@ -78,6 +78,32 @@ test('resets scope', function(assert) {
...
+    <em>ipsum</em>
+    <em style="display:none">dolor</em>
+  `);
+
+  let page = create({
+    foo: isHidden('em', { multiple: true })
+  });
+
+  assert.ok(!page.foo);
+});
```
san650: We need to test that it throws an error when it matches more than one element but `multiple` flag is not set

test-support/page-object/properties/count.js
```diff
@@ -22,6 +22,7 @@ export function count(selector, options = {}) {
     isDescriptor: true,
 
     get() {
+      options.multiple = true;
       return findElement(this, selector, options).length;
```
san650: We can use `Ember.$.merge({}, options, { multiple: true })` here so we don't mutate client objects

juanazam:
@san650 I implemented most of the tweaks, there might still be room for improvement.

san650:
@juanazam :trophy: :trophy: :trophy: :tada: :tada: :beers:

PR #120, opened 2016-02-04
Change `.visitable` signature

### LINKED ISSUE #113: [1.0] Change `.visitable` method signature
Current API

```js
  visit: visitable('/path/:user_id')
});

page.visit({ user_id: 5 }, { myQueryParam: 'foo' });  // "/path/5?foo=bar"
```

The generated function receives two different objects, the first one is for the dynamic segments and the second is for query params.

The problem arise when you don't have dynamic segments but you want to use query params

```js
page.visit({}, { myQueryParam: 'foo' });  // "/path/?foo=bar"
```

So, I would like to change the API so the function generated by `.visitable` receives only one parameter when it's invoked.

```js
  visit: visitable('/path/:user_id')
});

page.visit({ user_id: 5, foo: 'bar' }); // "/path/5?foo=bar"
```

The idea is to extract all dynamic segments first from the parameter and use the rest of the key/values as query params.

Note that If you don't set the values for all the dynamic segments, then an exception will be thrown (this is the current behavior).

san650:
Instead of receiving two distinct object parameters (dynamic segments
and query params) now it receives only one.

The idea is to fill the dynamic segments first using the values from
the param that match the keys and then use the rest of the keys and
values as query params.

e.g.

```
var page = create({
  visit: visitable('/users/:user_id')
})

page.visit({ user_id: 1, expanded: true });

// is equivalent to

visit("/users/1?expanded=true")
```

Fixes #113

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# master
+
+- [Change] Change `.visitable` signature
+
 # v1.0.0-alpha.1
 
 - [Feature] Can import properties directly `import { create } from 'path/to/page-object';`
--- a/test-support/page-object/properties/visitable.js
+++ b/test-support/page-object/properties/visitable.js
@@ -1,17 +1,23 @@
 import Ember from 'ember';
 
+var { merge, $ } = Ember;
+
 function fillInDynamicSegments(path, params) {
   return path.split('/').map(function(segment) {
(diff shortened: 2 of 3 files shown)


san650:
@juanazam can you review it?

jeradg:
:+1:

PR #121, opened 2016-02-04
Change `.clickOnText` to match current element

### LINKED ISSUE #114: [1.0] Change `.clickOnText` to match current element
Current API implementation only considers the following scenario

```html
<div class="modal">
  ...
  <button>Save</button> <button>Cancel</button>
</div>
```

```js
  clickOn: clickOnText('.modal')
});

page.clickOn('Save');
```
This triggers a click event on the Save button element.

Now consider the following scenario (same HTML)

```js
  clickOn: clickOnText('button')
});

page.clickOn('Save');
```

This won't work because the current implementation looks for elements inside button that contain the text but not the element itself.

We want both scenarios to work, so the idea is to first look for child elements and if none it's found, then try with the element itself.

san650:
Now the property can match both, current element or child elements

Fixes #114

### DIFF
--- a/test-support/page-object/properties/click-on-text.js
+++ b/test-support/page-object/properties/click-on-text.js
@@ -1,7 +1,27 @@
 import Ember from 'ember';
 import { buildSelector } from '../helpers';
 
+/* global wait, find, click */
+
 var { merge } = Ember;
+
+function findChildElement(tree, selector, textToClick, options) {
+  // Suppose that we have something like `<form><button>Submit</button></form>`
+  // In this case <form> and <button> elements contains "Submit" text, so, we'll
+  // want to __always__ click on the __last__ element that contains the text.
+  var selctorWithSpace = (selector || '') + ' ';
+  var fullSelector = buildSelector(tree, selctorWithSpace, merge({ contains: textToClick, last: true }, options));
+
+  if (find(fullSelector).length) {
+    return fullSelector;
+  }
+}
+
+function findElement(tree, selector, textToClick, options) {
+  var fullSelector = buildSelector(tree, selector, merge({ contains: textToClick }, options));
+
+  return fullSelector;
+}
 
 /**
  * Creates an action to click an element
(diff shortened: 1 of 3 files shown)


san650:
@juanazam can you review it?

test-support/page-object/properties/click-on-text.js
```diff
@@ -1,8 +1,28 @@
...
 
+/* global wait, find, click */
+
 var { merge } = Ember;
 
+function findChildElement(tree, selector, textToClick, options) {
+  // Suppose that we have something like `<form><button>Submit</button></form>`
+  // In this case <form> and <button> elements contains "Submit" text, so, we'll
+  // want to __always__ click on the __last__ element that contains the text.
+  var selctorWithSpace = (selector || '') + ' ';
```
juanazam: ` ${selector || ''}` seems more readable what do you think?

test-support/page-object/properties/click-on-text.js
```diff
@@ -1,8 +1,28 @@
...
+  var fullSelector = buildSelector(tree, selctorWithSpace, merge({ contains: textToClick, last: true }, options));
```
juanazam: This line seems to long, also are we using `var` instead of `let` for some reason?

test-support/page-object/properties/click-on-text.js
```diff
@@ -25,17 +45,11 @@ export function clickOnText(selector, options = {}) {
...
-      var selctorWithSpace = (selector || '') + ' ';
-      var fullSelector = buildSelector(
-        this,
-        selctorWithSpace,
-        merge({ contains: textToClick, last: true }, options));
-
-      /* global click */
-      click(fullSelector);
+      wait().then(() => {
+        var actualSelector = findChildElement(this, selector, textToClick, options) || findElement(this, selector, textToClick, options);
```
juanazam: Also too long :smile:

test-support/page-object/properties/click-on-text.js
juanazam: Also, isn't this equivalent to \` ${selector}\` when selector is undefined?

san650: I prefer to avoid expressions inside interpolated strings, I think it's more error prone. Also, `${undefined}` returns `undefined` :smile: JS is awesome!

test-support/page-object/properties/click-on-text.js
san650: I prefer `var` everywhere, I think it's less confusing. Maybe we should adhere to a code convention in the future.

test-support/page-object/properties/click-on-text.js
san650: haha not that long!

test-support/page-object/properties/click-on-text.js
juanazam: JS rules haha, I still preffer \` ${selector || ''}\` appending string with `+` reminds me of dark times haha.

PR #129, opened 2016-02-16
'npm run docs' command and restructuring of `page-object` subdirectories

jeradg:
Adds a script `npm run docs` to automatically generate GitHub Pages markdown from jsdoc comments.

Uses the `documentation` package to parse jsdoc comments. Output from `documentation` is massaged by `docs.js` and copied to the `gh-pages` branch.

PR also replaces the single `test-support/page-object/properties` subdirectory with separate subdirectories for `actions`, `predicates`, and `queries`.


juanazam:
@jeradg great work!

The good thing with `ember-cli-json` is that the user was able to generate the docs and run `ember s` to access a site with just the jsdoc documentation, but we can think of alternatives later.

I'll try this and merge it.

san650:
Great work @jeradg!

Issue #106, opened 2016-01-27
clickOnText action doesn't work


martinciu:
`clickOnText` creates selector with a white space before `:contains` for example: 

```javascript
# page object
clickAricle: clickOnText('h5');
```

```javascript
# test
page.clickArticle("Article 1");
```

generates selector:
`h5 :contains('Article 1'):last`
 but it should be: 
`h5:contains('Article 1'):last`

Is this test correct: https://github.com/san650/ember-cli-page-object/blob/master/tests/unit/actions/click-on-text-test.js#L10 ?

Selector like this doesn't work for me if I try it in the browser console.

san650:
Hi @martinciu, Actually, the intention is to use the selector to scope the container in which to look for elements that contain that phrase. e.g. if you have a modal with buttons, you would do something like

```js
clickOn: clickOnText('.modal')
```

and then

```js
page.modal.clickOn('Submit');
```

The idea is that you scope the search by the element that matches the selector. If you already know the exact element you want to click on then you should use the `clickable` attribute.

Saying that, maybe there's an use case that I'm not seeing, please let me know.

Looking at the documentation [here](http://ember-cli-page-object.js.org/api/actions/#clickontext) I noticed that it actually says something similar to what you say :sweat_smile: , so we need to fix the documentation because it never worked that way.

martinciu:
Hi @san650 , thanks for your reply.

This is my use case: I have several `h5` tags and I want to click only a few of them. For example:

```html
<h5>Article 1</h5>
<h5>Article 2</h5>
<h5>Article 3</h5>
<h5>Article 4</h5>
```

And I want to click Article 1 & 3. 

`clickable` won't work for this use case, will it?

san650:
@martinciu no, it won't work for that use case.

You can create a [custom helper](http://ember-cli-page-object.js.org/api/components/#customhelper) though

```js
var clickTitle = customHelper(function(selector) {
  return function(text) {
    console.log(`Selector: ${selector}:contains("${text}")`);
    return click(`${selector}:contains("${text}")`);
  };
});

var page = PageObject.create({
  // ...
  title: clickTitle('h5')
});

// ... and in the test ....
page.visit().title('Log in');
```

Please, let me know if that works for you.

mitchlloyd:
@san650 I just started using this library and I'm very impressed!

[...]

PR #137, opened 2016-02-24
Update example to match new API

mitchlloyd:
That little space makes all the difference when it comes to a CSS selector.

To close https://github.com/san650/ember-cli-page-object/issues/106

### DIFF
--- a/docs/v1.0.x/api/actions.md
+++ b/docs/v1.0.x/api/actions.md
@@ -60,7 +60,7 @@ const page = PageObject.create({
   clickOnButton: PageObject.clickOnText('button', { scope: '.scope' })
 });
 
-// queries the DOM with selector '.scope fieldset :contains("Lorem"):last'
+// queries the DOM with selector '.scope fieldset:contains("Lorem"):last'
 page.clickOnFieldset('Lorem');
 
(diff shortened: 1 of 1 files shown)


juanazam:
@mitchlloyd if you check over here: https://github.com/san650/ember-cli-page-object/blob/master/tests/unit/actions/click-on-text-test.js#L63, you can see that the example behaves as it's described on the docs.

mitchlloyd:
@juanazam Ok, I'm so confused. How is it that what appears to be the same syntax for `button` and `fieldset` produces different results?

juanazam:
The difference is that `fieldset` contains a `button` so the property doesn't behave the same way because of the way is implemented. This commit explains the change better that I can https://github.com/san650/ember-cli-page-object/commit/882d07f9010acc7b4bf8cb121cec8ec451eaeab4. In particular this line: https://github.com/san650/ember-cli-page-object/commit/882d07f9010acc7b4bf8cb121cec8ec451eaeab4#diff-d07a72752ffa301b4c0fdc55f7fa2a75R9

If you want you can jump in the `ember-cli-page-object` channel on the ember community slack.

PR #151, opened 2016-03-22
Update ember-cli to v2.4.2

san650:


jeradg:
@san650: Do we need ember-data as a dependency?

san650:
@jeradg you're right, I'll remove it

san650:
@jeradg You're right, we don't need ember-data

PR #155, opened 2016-03-23
Migrating from `test-support/` into `addon/`

lolmaus:
As coordinated with @jeradg in Slack, addon files have been moved into the `addon/` folder.

The import path is now

```js
import PageObject, {create} from 'ember-cli-page-object';
```

## Reasoning

1. Storing all addon source files under the `addon/` folder is a practice encouraged and enforced by Ember CLI. The default Ember CLI addon boilerplate does that, and files under `app/` are merely proxies to real ones under `addon/`.
2. For less experienced Ember users, importing from `my-app/tests/page-object` is frustrating. Essentially, it's importing a file that doesn't exist. I feel uncomfortable every time I do that, even though I understand the trick behind the magic.
3. To some, importing from `'ember-cli-page-object'` might look slightly bulkier than `'../../page-object'`, but the benefit of explicitness and clarity outweighs a few extra characters.
4. When PageObject code is stored under `test-support/`, third-party addon creators have no other way to expose their page objects than via `test-support/`, closing the vicious circle.
5. If an Ember rookie gets curious what can be imported from the addon, it's now obvious where to look.
6. The user now has control of whether to include `ember-cli-page-object` modules into production builds or not. **By default they are excluded.** If some weird use case requires them to be present in production, this can be done using an environment option. FYI, the difference `ember-cli-page-object`+`ceibo` make is 13 KiB (minified, non-gzipped).

## What to pay attention to in this PR

1. The main file is now `addon/index.js` (ex `test-support/page-object.js`). It can be imported simply with `from 'ember-cli-page-object'`.
2. I've revamped the structure of the main file. The new structure is drastically more concise and readable.
3. The old main file `test-support/page-object.js` still exists. It now proxies to `addon/`, so that existing imports won't break.
4. The old main file emits a deprecation warning. I tried using `Ember.deprecate()`, but it didn't work, probably because it's used on the top level. I used a simple `console.warn()` instead, mimicking an Ember deprecation.
5. I intentionally did not update the tests (except for fixing numerous JSHint errors). That is to demonstrate that old imports still work. Can update imports in tests any time.
6. I've updated paths for doc and blueprint generators. I tested generating blueprints but not docs.
7. I took liberty to update blueprint content. Now using named imports instead of destructuring from the `PageObject` proxy object. For simplicity and consistency, now using `create()` instead of `PageObject.create()`.
8. I failed to find any usages of the old path either in the Readme or the docs. 
9. I borrowed modules-exclusion-from-production logic from ember-cli-mirage. It even includes a Node-driven test.
10. I had to pull #151 (Update ember-cli to v2.4.2). For a clean diff, merge #151 and come back here.
11. After I merged #151, some tests broke. I fixed them. This includes a really [weird thing](https://github.com/lolmaus/ember-cli-page-object/blob/272e8f66b4e1750fbce329be2c71fb1a4017d187/tests/integration/component-integration-test-support/user-list-test.js#L89). But it's a separate issue worth investigating outside this PR.

BANAN**A**!


csantero:
Doesn't this mean that all the new testing-related files will be included in production builds?

san650:
@lolmaus thanks for taking the time to contribute.

I have the same concerns that @csantero has. When I created the addon I was using the structure you're proposing now @lolmaus but I had to change it because everything was included in the production build. But test-support/ is excluded from production build. Maybe this changed lately and I'm not aware of it...

lolmaus:
Dang it! I was absolutely sure that only imported files make it into the production build, starting at `app.js`.

I've just measured, `ember-cli-page-object` code in `vendor.js` is only 2957 bytes gzipped. But the very fact of it being included ruins the purpose.

I'll look into a way to exclude it.

lolmaus:
6.) The user now has control of whether to include `ember-cli-page-object` modules into production builds. **By default they are excluded.** If some weird use case requires them to be present in production, this can be done using an environment option.


## What to pay attention to in this PR

9.) I borrowed modules-exclusion-from-production logic from ember-cli-mirage. It even includes a Node-driven test.
10.) I had to pull #151 (Update ember-cli to v2.4.2). For a clean diff, merge #151 and come back here.
11.) #151 broke some tests. I fixed them. This includes a really [weird thing](https://github.com/lolmaus/ember-cli-page-object/blob/272e8f66b4e1750fbce329be2c71fb1a4017d187/tests/integration/component-integration-test-support/user-list-test.js#L89). But it's a separate issue worth investigating outside this PR.

lolmaus:
Fixed tests. :green_heart:

csantero:
@lolmaus Can the imports be made to only appear when running in test mode only? [In ember-sinon](https://github.com/csantero/ember-sinon/blob/master/index.js) I inspect `app.tests` to determine whether to include test assets.

san650:
Please, do not merge this until we discuss about it. Give me some days to investigate some concerns I have about this PR, we can discuss them on the slack channel and on this PR.

I just don't want to merge this in without taking care that everything will work just fine. Also, we just released a major version that was backward incompatible and this PR would make us create another release that is -- again -- backwards incompatible.

cc/ @lolmaus @jeradg @juanazam

lolmaus:
@san650,

I've been waiting for you in Slack with no luck.


> I just don't want to merge this in without taking care that everything will work just fine.

I've used it in our commercial app and it works fine. Both new and old paths.


> we just released a major version that was backward incompatible and this PR would make us create another release that is -- again -- backwards incompatible

This PR does not break anything at all. I intentionally did not update tests -- to show that the old paths keep working.

The breaking change -- removing the `test-support/page-object.js` proxy file -- can be postponed up to version 2.0.0. That's what Ember calls "stability without stagnation".


cc/ @jeradg @juanazam

san650:
Hi @lolmaus, sorry I didn't answer your message in slack. I've been a bit busy these days. I'll have some time this weekend to look at the PR.

Thanks for the clarifications about the backward compatibility concerns I have, still I would like to do some testing on my side to ensure everything works fine, this is a big PR I and it does change a lot of the internals so I want to be extra careful.

Let me do the review this weekend and I'll reach you here if any questions/concerns arises. Does this sound fine with you?

lolmaus:
@san650 Absolutely! Thank you for taking time and effort looking into the PR.

Don't hesitate to reach me on Slack for any clarifications or requesting for changes.

jeradg:
@san650 @lolmaus: Side note: The `docs.js` documentation generator will need to be updated if we proceed with this PR. (Happy to take that on.)

san650:
@lolmaus First, let me say thank you for the PR, I agree with your reasoning about why the changes make sense and I'm glad you took the time to make the migration.

> 3. To some, importing from 'ember-cli-page-object' might look slightly bulkier than '../../page-object', but the benefit of explicitness and clarity outweighs a few extra characters.

What do you think about using `page-object` instead of using the full name of the NPM package?

So user would have to do `import PageObject from 'page-object';` which I think is pretty neat.

@lolmaus, @juanazam, @jeragd, @csantero thoughts on this?

> 6. The user now has control of whether to include ember-cli-page-object modules into production builds or not. By default they are excluded. If some weird use case requires them to be present in production, this can be done using an environment option. FYI, the difference ember-cli-page-object+ceibo make is 13 KiB (minified, non-gzipped).

Why would anyone would want to add the addon to a production build? Maybe we can wait for someone to open an issue asking for this feature before adding it to the addon, this way we don't have features that nobody uses. Is this OK with you?

> 4. The old main file emits a deprecation warning. I tried using Ember.deprecate(), but it didn't work, probably because it's used on the top level. I used a simple console.warn() instead, mimicking an Ember deprecation.

This is weird, Is this an ember bug?

> 5. I intentionally did not update the tests (except for fixing numerous JSHint errors). That is to demonstrate that old imports still work. Can update imports in tests any time.

We can address this in a following PR (I can help with this if you want). I would like to use the `import ... from 'page-object';` notation everywhere and have a specific test that uses the old import to make sure we don't break it in the near future. Again, this can be done in a follow up PR.

> 6. I took liberty to update blueprint content. Now using named imports instead of destructuring from the PageObject proxy object. For simplicity and consistency, now using create() instead of PageObject.create().

:heart:

> 10. I had to pull #151 (Update ember-cli to v2.4.2). For a clean diff, merge #151 and come back here.

I've merged this PR into `master`

> 11. After I merged #151, some tests broke. I fixed them. This includes a really weird thing. But it's a separate issue worth investigating outside this PR.

`master` with #151 is not failing and I didn't include that fix... Let's see if rebasing with master don't make that problem to go away



Wow, that was a lot of work that you've done, I'm migrating a couple of small projects I have with different versions of ember to see if there's any issue with the changes you've made. And I'm looking at the code so I might add some extra comments.

san650:
@jeradg 

> @san650 @lolmaus: Side note: The docs.js documentation generator will need to be updated if we proceed with this PR. (Happy to take that on.)

We can do this as a follow up PR for the `gh-pages` branch, right? If so, let's wait to merge this PR so we don't do extra work :)

jeradg:
@san650: Yes, that's what I'm thinking.

blueprints/page-object/files/tests/pages/__name__.js
```diff
@@ -1,9 +1,8 @@
-import PageObject from '<%= pageObjectsRoot %>';
-
-let {
+import {
+  create,
   visitable
-} = PageObject;
+} from '<%= pageObjectsRoot %>';
```
san650: Let's remove the `pageObjectsRoot` helper.

lib/blueprints-helpers.js
```diff
@@ -1,11 +1,7 @@
-var stringUtils = require('ember-cli-string-utils');
-
 module.exports = {
   locals: function(options) {
-    var folder = stringUtils.dasherize(options.project.name()) + '/tests/page-object';
-
     return {
-      pageObjectsRoot: folder,
+      pageObjectsRoot: 'ember-cli-page-object'
```
san650: We don't need this file anymore.

csantero:
> Why would anyone would want to add the addon to a production build? Maybe we can wait for someone to open an issue asking for this feature before adding it to the addon, this way we don't have features that nobody uses. Is this OK with you?

@san650 I agree that it doesn't make sense that a production build would want these modules available, and furthermore I believe that they don't belong in a non-testing development build either. Luckily this should be an easy change: in the `_shouldIncludeFiles` method in `index.js`, inspect `app.tests` instead of the `app.env` configuration.

san650:
@csantero thanks for the tip.

lolmaus:
> @san650:
> 
> @lolmaus First, let me say thank you for the PR, I agree with your reasoning about why the changes make sense and I'm glad you took the time to make the migration.
> 
> ***
> 
> Wow, that was a lot of work that you've done (...)

My pleasure!

*** 

> @san650:
> 
>> .3 To some, importing from 'ember-cli-page-object' might look slightly bulkier than '../../page-object', but the benefit of explicitness and clarity outweighs a few extra characters.
> 
> What do you think about using page-object instead of using the full name of the NPM package?
> 
> So users would have to do import PageObject from 'page-object'; which I think is pretty neat.

Done! https://github.com/lolmaus/ember-cli-page-object/commit/b615cf865b823a5316aea8b0c66d45bfde2ebb4c

***

> @san650:
> 
>> .6 The user now has control of whether to include ember-cli-page-object modules into production builds or not. By default they are excluded. If some weird use case requires them to be present in production, this can be done using an environment option. FYI, the difference ember-cli-page-object+ceibo make is 13 KiB (minified, non-gzipped).
> 
> Why would anyone would want to add the addon to a production build? Maybe we can wait for someone to open an issue asking for this feature before adding it to the addon, this way we don't have features that nobody uses. Is this OK with you?

Done! https://github.com/lolmaus/ember-cli-page-object/commit/78127604ab7e482d3f04bb19f09fcf935de372eb

***

> @san650:
> 
>> .4 The old main file emits a deprecation warning. I tried using Ember.deprecate(), but it didn't work, probably because it's used on the top level. I used a simple console.warn() instead, mimicking an Ember deprecation.
> 
> This is weird, Is this an ember bug?

No idea. :(

***

> @san650:
> 
>> .5 I intentionally did not update the tests (except for fixing numerous JSHint errors). That is to demonstrate that old imports still work. Can update imports in tests any time.
> 
> We can address this in a following PR (I can help with this if you want). I would like to use the import ... from 'page-object'; notation everywhere and have a specific test that uses the old import to make sure we don't break it in the near future. Again, this can be done in a follow up PR.

Cool!

***

> @san650:
> 
> Let's remove the `pageObjectsRoot` helper.

Done! https://github.com/lolmaus/ember-cli-page-object/commit/7207c79e704f6b099fbe76c9eac6d122a0ae7f6f

***

> @csantero:
>
>> @san650:
>> 
>> Why would anyone would want to add the addon to a production build? Maybe we can wait for someone to open an issue asking for this feature before adding it to the addon, this way we don't have features that nobody uses. Is this OK with you?
> 
> @san650 I agree that it doesn't make sense that a production build would want these modules available, and furthermore I believe that they don't belong in a non-testing development build either. Luckily this should be an easy change: in the _shouldIncludeFiles method in index.js, inspect app.tests instead of the app.env configuration.

If you remove page object modules from development builds, it will be impossible to start `ember s` in order to work with http://localhost:4200 and at the same time see tests status at http://localhost:4200/tests .

This is my standard workflow, and I believe lots of other developers use it to. Removing page object modules from development builds will force us to run two servers in parallel: `ember s` and `ember t -s`. Moar compiling, moar file watching! :trollface:

jeradg:
@san650 @lolmaus: Is there a standard for what to refer to the addon as when importing? I.e., are we going against a best practice by importing from `page-object` instead of `ember-cli-page-object`?

lolmaus:
> @san650 @lolmaus: Is there a standard for what to refer to the addon as when importing? I.e., are we going against a best practice by importing from page-object instead of ember-cli-page-object?

I assume no and no. The typical scenario is the npm package and addon's main module to be named identically, but I'm not aware of a convention enforcing that. Having `ember-cli-` in an npm package name makes total sense, but repeating all those `ember-cli-` within an Ember CLI app is pointless.

@stefanpenner and @rwjblue could give an authoritative answer.

rwjblue:
We added a new tree a few versions back, that works like test-support (it is included only in testing and not main vendor.js/app.js) AND allows importing from the addons namespace. 

Placing a file on disk in the addon at 'addon-test-support/foo.js' allows imports like:

    import foo from 'ember-cli-page-object/tests/foo';

lolmaus:
T_T

rwjblue:
> I assume no and no. The typical scenario is the npm package and addon's main module to be named identically, but I'm not aware of a convention enforcing that. Having ember-cli- in an npm package name makes total sense, but repeating all those ember-cli- within an Ember CLI app is pointless.

I disagree. Having an import path that does not match your package namespace is very confusing ("where does this page-object thing come from?"), and goes against normal expectations in the overall JS ecosystem (npm ENFORCES the package name to be the name that you use with 'require'). 

I see no reason to make usage of the addon more confusing...

jeradg:
Thanks @rwjblue.

@lolmaus and @san650, I'd say let's go with the conventional import name, and see if we can get the addon-test-support working. The import path is longer, but a) we have generators that mean people don't need to type the path at all, and b) we can avoid extra configuration to exclude the addon from non-test builds.

lolmaus:
@rwjblue What about this new tree? How is it called? Is it documented? Where we can read about it?

rwjblue:
It is documented in ember-cli:

* the addon model
* ADDON_HOOKS.md
* CHANGELOG.md

If those docs don't have enough detail, I would love a PR fleshing things out more.

lolmaus:
@rwjblue Thx!

@san650 @jeradg I will revert module name change and update the PR to use the feature proposed by @rwjblue.

Probably tomorrow.

san650:
@rwjblue thanks for the heads up about `addon-test-support`.

@lolmaus @jeradg I'm not sure if using `addon-test-support` breaks compatibility of the addon with older versions of ember-cli.

I have a pending task which is to test this addon against older versions of ember-cli to see what versions we support right now. And if you ask me, I would like to support at least ember-cli 1.13+

If we use `addon-test-support` would we loose compatibility with older versions of ember-cli? Is the addon already incompatible with ember-cli 1.13? I can investigate about this tonight and see where we are at.

About the import path, I still like the `page-object` path, I don't think it's a big deal __but__ I'm ok with going with the best practice and use the full NPM package name instead.

jeradg:
@san650 @lolmaus: According to our ember-try tests, we have 1.13 compatibility before this PR.

I agree that we don't want to break compatibility so soon after releasing 1.x. So the real trick is going to be adding this new import path while still having the old one work. May or may not be possible.

san650:
@jeradg ember-try only test against different versions of ember and not ember-cli itself. I would like to know if the support of ember-cli 1.13 is broken or not.

lolmaus:
> Is the addon already incompatible with ember-cli 1.13?

Travis is green for all versions, including 1.12 and 1.13.

***

> I still like the page-object path, I don't think it's a big deal
>
> we don't want to break compatibility so soon after releasing 1.x

Good to merge then? :trollface:

jeradg:
@san650: I was using with ember-cli 1.13 until a few days ago and it was working fine... If that helps.

lolmaus:
1. Reverted module renaming, now using `ember-cli-page-object` as per @rwjblue's recommendation.
2. Tested both old and new import paths in an app using `ember-cli`@`1.13.15`.

Codacy is happy too: https://www.codacy.com/app/lolmaus/ember-cli-page-object/commit?bid=3200493&cid=31001878

lolmaus:
Bump?

jeradg:
@lolmaus: Hey! The team is all at EmberConf at the moment. Might take us a couple of days to review.

lolmaus:
image

README.md
```diff
@@ -54,7 +54,7 @@ $ ember generate page-object users
 
 installing
   create tests/pages/users.js
-```
+
```
san650: This change breaks the markdown

index.js
```diff
@@ -7,8 +7,24 @@ module.exports = {
   included: function(app) {
     this._super.included(app);
 
-    if (app.env === 'test' || app.env === 'development') {
+    this.app = app;
+    this.addonConfig = this.app.project.config(app.env)['ember-cli-page-object'] || {};
+    this.addonBuildConfig = this.app.options['ember-cli-page-object'] || {};
```
san650: Do we need to read the configuration? We're not adding new options for the add-on.

test-support/page-object.js
```diff
@@ -58,3 +58,5 @@ export default {
   value,
   visitable
 };
+
+console.warn("DEPRECATION: Importing ember-cli-page-object from your app's `tests/` folder has been deprecated. Instead, import from an absolute path like 'ember-cli-page-object', 'ember-cli-page-object/create', etc.");
```
san650: Don't print a deprecation warning for now, let's add the feature and see how it goes.

tests/fixtures/config/environment-production-enabled.js
```diff
@@ -0,0 +1,50 @@
...
+
+  if (environment === 'production') {
+    ENV['ember-cli-page-object'] = {
+      enabledInProd: true
+    };
+
+  }
+
+  return ENV;
+};
```
san650: We don't need this because we don't have options to test now.

package.json
```diff
@@ -12,8 +12,9 @@
     "build": "ember build",
     "preversion": "git clean -df && npm test",
     "start": "ember server",
-    "test": "ember try:testall",
-    "docs": "node ./docs"
+    "test": "ember try:testall && mocha tests/**/*-test-node.js",
+    "docs": "node ./docs",
+    "mocha": "mocha tests/**/*-test-node.js"
```
san650: We don't need node tests for this feature now that we don't have the `enabledInProd` option

san650: Looking at the node tests I realized that we DO need this test :smile:

tests/unit/node/addon-tree-test-node.js
```diff
@@ -0,0 +1,70 @@
...
+    ['development', 'test'].forEach(function(environment) {
+      it('returns a tree in ' + environment + ' environment by default', function() {
+        process.env.EMBER_ENV = environment;
+        var addonTree = getAddon().treeFor(name);
+
+        expect(addonTree._inputNodes.length).to.not.equal(0);
+      });
+    });
+
+    it('returns a tree in production environment when enabled is specified', function() {
```
san650: We don't need this test

san650:
Hi @lolmaus, let's try to merge this PR. I added some comments. I think the PR will be ready after this comments are addressed. Also, rebase with master and squash the commits if you can.

Apart from that, update the AUTHORS file.

Thanks for this incredible work and sorry for the delay on our part to merge it.

:beers:

package.json
jeradg: This requires a globally installed Mocha. If we need to run Mocha tests, we should add Mocha to the package.json and run the local version.

rwjblue: Mocha is added below in `devDependencies`.  `scripts` are ran with the `node_modules/.bin` directory in `$PATH`, so all binaries added by deps/devDeps are available.

jeradg: :thumbs up:

jeradg: Argh autocorrect. (That's supposed to be a :+1:)

lolmaus:
Ready for merge. Please someone give it a try in a project for a second look before merging.

I've added blank lines to tests because *it friggin' fixes them*. I've compared compiled JS output with or without extra line -- and it's totally different. I have no idea what's going on, but it's definitely not related to this PR. Some weird Babel issue.

san650:
@lolmaus Thanks, I'll take a look at it today

san650:
Thanks for contributing!

lolmaus:
@san650 

image

juanazam:
@lolmaus you can finally update to `v1.12.0` and get this :clap: thanks again!

PR #164, opened 2016-04-18
Make collections easier to enumerate

dfreeman:
This PR implements the changes discussed in https://github.com/san650/ember-cli-page-object/issues/54. @san650 it also includes the delegation of array methods we talked about, since it was easy to add them while I was in there. If you'd prefer to have that addressed in a separate PR, I'm happy to pull them out.

With this change, the following properties are now added to collection objects on creation if they haven't already been defined:
- a `toArray()` method that returns an array containing the elements of the collection
- a `[Symbol.iterator]()` method which yields each item in the collection sequentially (only if `Symbol.iterator` is present in the current environment)
- `map`, `mapBy`, `filter`, and `filterBy` methods, all of which just delegate to the methods of the same name on the result of `toArray()`.

### DIFF
--- a/addon/collection.js
+++ b/addon/collection.js
@@ -4,6 +4,7 @@ import { create } from './create';
 import { count } from './queries/count';
 
 const mergeFunction = Ember.assign || Ember.merge;
+const arrayDelegateMethods = ['map', 'filter', 'mapBy', 'filterBy'];
 
 function merge(target, ...objects) {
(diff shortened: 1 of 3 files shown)


san650:
Thanks for contributing!

juanazam:
@dfreeman this fix is included on `v1.12.0`, cheers and thanks!

san650:
@dfreeman @juanazam it's `v1.2.0` 😄

PR #165, opened 2016-04-19
Uncomment and update default properties acceptance test

juanazam:


juanazam:
@san650 @jeradg small PR to re add those tests, please review these changes

jeradg:
@juanazam: Looks OK to me. I'm a little fuzzy on how exactly the calculator works though (i.e., the controller logic). That's probably not super important though.

juanazam:
@jeradg I agree, also we have duplication with the controller and the component, we should unify that in the future. Can you merge this?

PR #167, opened 2016-04-22
Add .property property

### LINKED ISSUE #112: [Feature] .property query
I find myself having to write this sort of thing a lot:

```js
assert.equal(page.foo.attribute('disabled'), 'disabled');
```

It would be good to have a `.property` query to check for element boolean properties, ex.:

```js
assert.ok(page.foo.property('disabled'));
```

san650:
Fixes #112

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # master
 
+- [Feature] Add `property` property
+
+# v1.2.0
+
 - [Feature] Add option to text property to avoid text normalization
 - [Feature] Add `toArray` method to collections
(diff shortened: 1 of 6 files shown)


san650:
@juanazam @jeradg ready for code review

jeradg:
@san650: There's a typo in your commit. In `addon/index.js`:

```
import { proeprty }    from './queries/property';         export { property };
```

jeradg:
@san650: Probably means that we need to test that the exports from `addon/index.js` actually work.

san650:
I'll add a test for it

PR #168, opened 2016-04-23
Add test container option

san650:
Add support for targeting elements outside ember application container. Useful for testing components that append elements to the body in tests.

```js
var page = create({
  title: text('h1', { testContainer: 'body' })
});

assert.equal(page.title, 'foo');
```

This PR adds this option to every property for acceptance and component integration tests


san650:
cc/ @juanazam @jeradg

jeradg:
@san650: This is a great idea!

One question: What's the reasoning for deleting those two integration tests? I'm worried that we now don't have test coverage ensuring that page objects will work correctly with component tests.

san650:
@jeradg I didn't take a deep look at those tests but my impression was that those two integration tests weren't testing anything new (i.e. the other integration tests cover the same cases).

If some tests cases are missing I can add those to the "calculator" tests.

jeradg:
@san650: As long as we have coverage of all of the properties/actions for components, it doesn't matter to me where they are. I originally copied the acceptance tests and reworked them into component tests. Makes sense to me that we would have the same coverage in both contexts.

san650:
I'll take a closer look at these tests and ensure we're not losing coverage. Thanks for the heads up!

san650:
@jeradg I've just checked the tests and I'm pretty sure all cases that were tested on the deleted tests are also covered in the existent ones.

Let me know if there's any change you would like to make to this PR.

jeradg:
@san650: Looks good to me. Look forward to trying it out.

san650:
@jeradg please merge it. Let's ignore the Codacy issue for now 😛

jeradg:
Yeah, Codacy is being completely useless lately.

PR #171, opened 2016-04-24
Add focus property

juanazam:
This pull request adds the `focus` property and it also includes it by default on every component.

```js
  scope: '#page',
  name: {
    scope: 'input'
  }
};

page.name.focus(); // focuses on "#page input"
```

### DIFF
--- a/addon/actions/focusable.js
+++ b/addon/actions/focusable.js
@@ -1,0 +1,94 @@
+import Ember from 'ember';
+import { simpleFindElementWithAssert, buildSelector, getContext } from '../helpers';
+
+function focusableInternal(tree, selector, options, context) {
+  var fullSelector = buildSelector(tree, selector, options);
+
+  // Run this to validate if the element exists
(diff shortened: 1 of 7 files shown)


juanazam:
@san650 @jeradg can any of you guys take a look at this?

san650:
Why `focusable` and not a `triggerEvent` wrapper? Creating a `triggerEvent` gives us a lot more options with less properties.

```js
   focus: triggerableEvent('focus', '.foo')
});
```

@juanazam What do you think? I don't want to add one property for each possible event.

YoranBrondsema:
@juanazam @san650 I ran into a similar issue where I needed the [native focus() method](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus). The reason is that `triggerEvent` doesn't set the `document.activeElement` property. So my understanding is that their semantics are slightly different. I resorted to creating a custom `focusable` action but having this PR merged would make it easier.

PR #172, opened 2016-04-26
Add triggerEvent property

juanazam:
This pull request adds the focus property and it also includes it by default on every component.

```js
  scope: '#page',
  
  focusOnName: triggerable('focus', 'input')
};

page.focusOnName(); // focuses on "#page input"
```

### DIFF
--- a/addon/actions/triggerable.js
+++ b/addon/actions/triggerable.js
@@ -1,0 +1,96 @@
+import Ember from 'ember';
+import { simpleFindElementWithAssert, buildSelector, getContext } from '../helpers';
+
+function triggerableInternal(tree, event, selector, options, context) {
+  var fullSelector = buildSelector(tree, selector, options);
+
+  // Run this to validate if the element exists
+  simpleFindElementWithAssert(tree, fullSelector, options)
+
+  if (context) {
+    if (options.testContainer) {
+      Ember.$(fullSelector, options.testContainer).trigger(event);
+    } else {
+      context.$(fullSelector).trigger(event);
+    }
+  } else {
+    /* global triggerEvent */
+    triggerEvent(fullSelector, options.testContainer, event);
+  }
+}
+
+/**
+ *
+ * Triggers event on element matched by selector.
+ *
+ * @example
+ *
+ * // <input class="name">
+ * // <input class="email">
+ *
+ * const page = PageObject.create({
+ *   focus: triggerable('focus', '.name')
+ * });
+ *
+ * // focuses on element with selector '.name'
+ * page.focus();
+ *
+ * @example
+ *
+ * // <div class="scope">
+ * //   <input class="name">
+ * // </div>
+ * // <input class="email">
+ *
+ * const page = PageObject.create({
+ *   focus: triggerable('focus', '.name', { scope: '.scope' })
+ * });
+ *
+ * // focuses on element with selector '.scope .name'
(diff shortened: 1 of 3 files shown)


juanazam:
@san650 @jeradg guys, can you please review it? :smile:

jeradg:
@juanazam @san650: My instinct is that the name of the event should be the first argument to `triggerable`. I think that fits with the signature of the other properties.

Also, if only the event name is passed (no selector is passed), should the event be fired on the current scope?

jeradg:
I think I've answered my second question, since that's how buildSelector works with an undefined target selector.

juanazam:
@jeradg it will work with the inherited scope.

I kept the same params order than with the original `triggerEvent` helper provided by ember, but I can change it if both @san650 and you agree.

san650:
@juanazam the event name should be the first parameter, so the property can be used without a selector

juanazam:
@san650 @jeradg I changed the parameter order :smile:

addon/actions/triggerable.js
```diff
@@ -0,0 +1,96 @@
...
+    }
+  } else {
+    /* global triggerEvent */
+    triggerEvent(fullSelector, options.testContainer, event);
+  }
+}
+
+/**
+ *
+ * Focus on element matched by selector.
```
san650: This description is not longer valid

addon/actions/triggerable.js
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * @example
+ *
+ * // <input class="name">
+ * // <input class="email">
+ *
+ * const page = PageObject.create({
+ *   focusOnName: triggerable('focus'. '.name')
```
san650: change the dot for a comma `%s/'focus'./'focus',/`

san650: Maybe use just `focus` as the name of the prop.

addon/actions/triggerable.js
```diff
@@ -0,0 +1,96 @@
...
+ * @example
+ *
+ * // <div class="scope">
+ * //   <input class="name">
+ * // </div>
+ * // <input class="email">
+ *
+ * const page = PageObject.create({
+ *   scope: '.scope',
+ *   continue: triggerable('focus', '.name')
```
san650: `s/continue/focusOnName/`

addon/actions/triggerable.js
```diff
@@ -0,0 +1,96 @@
...
+ * });
+ *
+ * // clicks on element with selector '.scope button.continue'
+ * page.focusOnName();
+ *
+ * @public
+ *
+ * @param {string} event - Event to be triggered
```
san650: `Event to trigger`... I know, I'm not the one to correct english grammar so what I'm suggesting could be wrong

juanazam: Both are correct :dancer:

tests/unit/actions/triggerable-test.js
```diff
@@ -0,0 +1,149 @@
+import { test } from 'qunit';
+import { moduleFor, fixture } from '../test-helper';
+import { create, triggerable } from '../../page-object';
+
+moduleFor('Unit | Propery | .triggerable');
```
san650: `s/Propery/Property/`

jeradg:
Nice work!

PR #175, opened 2016-04-27
Validate element visible on `clickable` attribute

jmbejar:
A fix for #6 🐼


san650:
Just so everyone is on the same page, I've just chatted with @jmbejar and decided to make this behaviour optional (by adding an option... `{ raiseIfHidden: true }`?) so the feature is backwards compatible and make it the default behaviour on 2.0.0.

Thanks @jmbejar for contributing.

juanazam:
@jmbejar @san650 great, we are talking about including this option for every lookup right?

san650:
released in v1.4.0. See https://github.com/san650/ember-cli-page-object/releases/tag/v1.4.0

PR #182, opened 2016-05-05
Add ember-suave

### LINKED ISSUE #180: Enforce ember style guides
It would be cool to start using something like [ember-suave](https://github.com/dockyard/ember-suave) to enforce a code style guide.

We can discuss here what rules we want to apply.

hidnasio:
Fixes #180 

Fix some indentation issues
Change var to let
Remove multiple variable declarations
Add a space after the switch keyword
Disable requireSpread rule in tests/helpers/module-for-acceptance.js file
Change const to let when used inside functions, const should be used only at module level
Add parenthesis to the arrow functions
Remove multiple line breaks
Disable requireCamelCaseOrUpperCase rule in test_throws_if_not_multiple. It should be camel case but it used all over the project.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,0 +1,4 @@
+{
+  "preset": "ember-suave",
+  "requireCamelCaseOrUpperCaseIdentifiers": false
+}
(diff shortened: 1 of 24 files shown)


san650:
Thanks for contributing!

san650:
released in v1.4.0. See https://github.com/san650/ember-cli-page-object/releases/tag/v1.4.0

PR #185, opened 2016-05-09
Add keyable property

juanazam:
Add keyable properties which maps to Ember's `keyEvent` testing helper.

``` js
page = create({
  foo: keyable('keypress', 13, 'span', { at: 3 })
});
```

### DIFF
--- a/addon/actions/keyable.js
+++ b/addon/actions/keyable.js
@@ -1,0 +1,102 @@
+import Ember from 'ember';
+import { simpleFindElementWithAssert, buildSelector, getContext } from '../helpers';
+
+function keyableInternal(tree, eventType, keyCode, selector, options, context) {
+  var eventOptions = options.eventProperties,
+    fullSelector = buildSelector(tree, selector, options);
+
+  delete options.eventProperties;
+
+  // Run this to validate if the element exists
+  simpleFindElementWithAssert(tree, fullSelector, options)
+
+  if (context) {
+    let event = Ember.$.Event(eventType, { keyCode });
+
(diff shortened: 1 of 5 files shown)


jeradg:
@juanazam: Nice!

I notice that there are no tests to make sure that the component-test version (with a `this` context) works. This is actually a bit of an issue with new features in general.

Want me to take a stab at these?

juanazam:
@jeradg I agree that it would be awesome to test this with the component-test version. I have mixed feelings about waiting for this to be merged before that is ready. I will give it a try these next days and see how easy it is. The tedious thing about this is to make the current test code work and also making sure the example makes sense with this property.

jeradg:
@juanazam @san650: 

I we could use some agreement on what standard to meet before merging a feature into master. Do we need to have both acceptance and integration test support? Or is just one enough, and full support can be added in a separate PR?

I personally feel like we should not release features that don't work for both component and acceptance tests.

We're advertising that the addon supports both acceptance and integration tests, but it's unclear how true that is over the last couple of releases. I'm happy to take the lead on the component test part. But I think in general supporting both should be part of any PR.

Thoughts?

juanazam:
@jeradg In an ideal world I would rather have support for both at the same time (including test for both cases). Right now doing tests for both cases is a bit difficult, it would be wise to spend some time re organizing the test code to make it easy to test new properties when creating them for both scenarios.

san650:
I'm not sure if these feature is needed. Also, the PR is pretty old so it should be updated. I'll close the PR for now and let's chat about the feature.

PR #194, opened 2016-06-03
Always import ember-template-compiler in dummy app

### LINKED ISSUE #193: Test failures when running tests from the dummy app
Tests that require ```ember-template-compiler.js``` fail on ```master``` when I run them from the dummy test app in the browser (as in ```localhost:4200/tests```)

Error message: 
```Cannot call `compile` without the template compiler loaded. Please load `ember-template-compiler.js` prior to calling `compile`.```

```
➙  ember -v
version: 2.4.2
node: 4.3.0
os: darwin x64
```

All tests pass if I run in test environment (as in ```ember test --server```)

I understand that most people run their tests the latter way, but In case anyone wants to run them from the browser, is there a downside to just including the ```ember-template-compiler.js``` in the ```ember-cli-build``` regardless of what environment we are running?

migbar:
Fixes #193

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -16,13 +16,10 @@ module.exports = function(defaults) {
     behave. You most likely want to be modifying `./index.js` or app's build file
   */
 
-  if (app.env === 'test') {
-    // Load ember-template-compiler for testing with Ember version <= 1.12;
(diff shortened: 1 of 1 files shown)


juanazam:
@migbar thanks for contributing, this is included on `v1.4.1`

PR #195, opened 2016-06-06
[EXPERIMENT] Improve test suite

san650:
THIS PR IS JUST AN EXPERIMENT, __DO NOT MERGE__

When testing a property we need to execute the same test for _acceptance_ mode and for _component integration_ mode.

This PR tries to reuse the same test to be run under both modes. We achieve this by adding three new testing helpers, `module`, `test` and `finish`.

The idea is that we create a group of tests for each property by calling `module` first and then calling `test` several times. When we finished declaring all tests for the property we call the `finish` helper which will create two copies of the module under the hood, one for _acceptance_ mode and another for _component integration_ mode.

You can see here two examples, one for an action property (with mocks) and another for a query property.

* https://github.com/san650/ember-cli-page-object/blob/acd03b6ccabd19335acd0c2398689e50f35954e2/tests/unit/actions/clickable-test.js
* https://github.com/san650/ember-cli-page-object/blob/acd03b6ccabd19335acd0c2398689e50f35954e2/tests/unit/queries/text-test.js

In order to achieve this I've extracted the differences for testing in _acceptance_ and _component integration_ modes into two adapters.

*  https://github.com/san650/ember-cli-page-object/blob/acd03b6ccabd19335acd0c2398689e50f35954e2/tests/helpers/acceptance-adapter.js
* https://github.com/san650/ember-cli-page-object/blob/acd03b6ccabd19335acd0c2398689e50f35954e2/tests/helpers/integration-adapter.js

The idea is to discuss the solution, see if we like it (or not) and look for improvement opportunities. __Don't hesitate to criticize the code or the solution!__

Just a small note, although I've implemented everything under `test/unit` folder, these tests are not unit test in the most formal definition. For _acceptance_ mode we're instantiating and booting an Ember application and for _component integration_ mode we're using the _component integration_ infrastructure. Personally I like to have a more lax definition of unit tests (test an atomic part of the app and ensure they run really fast 😄 ).


san650:
The build for ember-1-12 is failing in travis-ci. Let's not focus on that right now, let's focus on the design first.

PR #197, opened 2016-06-11
Don't include item and itemScope in collection component

### LINKED ISSUE #196: Iterating over collections broken
It seems I cannot iterate over collections without using an explicit index.

Look at the following example:
```js
      console.info('for-loop');
      for (let i=0; i < page.categories().count; i++) {
        let category = page.categories(i);
        console.log('meta', category.__meta.key);
        console.log('headline', category.headline);
      }

      console.info('for-of-loop');
      for (let category of page.categories()) {
        console.log('meta', category.__meta.key);
        console.log('headline', category.headline);
      }
```

This is the output I get:
image

The first for-loop is working fine, but there is something broken with the for-of loop: `__meta.key` is `undefined(0)`, and the `[data-test-selector="product-categories"]` selector is duplicated (this is from categories.scope, see below). The same happens when I change the for-of to a forEach construct:

```js
    page.categories().toArray().forEach((category) => {
        console.log('meta', category.__meta.key);
        console.log('headline', category.headline);
      });
```

My page object: (`selectElement` is just a helper function returning a DOM selector)

```js

export default create({
  visit: visitable('/de/shop/configurator/:id'),

  categories: collection({
    scope: selectElement('product-categories'),
    itemScope: 'section',
    item: {
      headline: text(selectElement('product-category-name')),
      products: collection({
        itemScope: selectElement('product'),
        item: {
          modelId: attribute('data-test-model'),
          name: text(selectElement('name'))
        }
      })
    }
  })

```

san650:
Fixes an error when trying to print the full collection into console or trying to use the `for ... of` construct.

The problem was that we were generating an invalid component `item` inside the collection component that was throwing an `Element not found` error when inspected. By removing this invalid element, the collection component can be inspected safely.

Fixes #196


juanazam:
👍

juanazam:
@san650 thanks for contibuting, this is included on `v1.4.1`

PR #208, opened 2016-07-01
Don't mutate definition objects

### LINKED ISSUE #207: `collection` now mutates its input
It looks as though [this commit](https://github.com/san650/ember-cli-page-object/commit/78fbb2d48370d35ba8f13fcd00ae1d36670a0e91) changed the `collection` helper so that it now deletes keys on its input hash, which broke several of our page objects. Up to 1.4.0, we had been sharing the POJOs for such definitions among related page objects to make it easy to refine e.g. the items in the collection for more specific use cases.

We could move to using factory functions instead so that every definition is guaranteed fresh input, but simple `...` object spread had been working quite nicely to this point, and it was surprising to me that any of the helpers would mutate their input.

[This twiddle](https://ember-twiddle.com/6f7689d10ec6877e484f9f0a0ab2b842?fileTreeShown=false&numColumns=3&openFiles=tests.integration.components.my-component-test.js%2Ctemplates.components.my-component.hbs%2Ctwiddle.json) shows an example that fails in 1.4.1 because of this change, but passes if you switch the dependency in `twiddle.json` to 1.4.0.

san650:
Fixes #207

### DIFF
--- a/addon/collection.js
+++ b/addon/collection.js
@@ -203,6 +203,8 @@ function iteratorMethod() {
  * @return {Descriptor}
  */
 export function collection(definition) {
+  definition = mergeFunction({}, definition);
+
   let item = {
     scope: definition.scope,
     itemScope: definition.itemScope,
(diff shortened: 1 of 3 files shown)


dfreeman:
@san650 Just confirmed this takes care of the issue we saw with 1.4.1. Thanks for taking care of it so quickly!

PR #214, opened 2016-07-07
add _site to npmignore

jakesjews:
This takes 13MB off the size of the installed package


san650:
Wow thanks for catching this!

Issue #226, opened 2016-08-02
Pause test on error option when running under QUnit


san650:
It would be great to have an option on QUnit to pause the test execution when the page object fails to find an element in the DOM.

We could add a checkbox in QUnit's index.html page so we could activate the feature when that option is selected.

<img width="473" alt="screen shot 2016-08-02 at 2 35 45 pm" src="https://cloud.githubusercontent.com/assets/1234607/17338607/70d8fb0c-58be-11e6-9bd7-3b9b219b0980.png">

The idea is that when you're working on a test, sometimes the selectors are incorrect and would be really useful if the tests stops with the rendered page and all the information about the missing page object property/selector in the javascript console.

jeradg:
@juanazam, are you interested? I'm happy to take this on.

fedekau:
@jeradg Hi! I started looking how to implement this as my first contribution to the project, If you are ok with that I would like to keep working on that :)

san650:
This will be implemented as a separated addon

PR #234, opened 2016-08-09
Pause test on error option when running under QUnit

fedekau:
This adds #226.

I am not sure how to test this right now. Any idea?

### DIFF
--- a/addon/-private/better-errors.js
+++ b/addon/-private/better-errors.js
@@ -1,7 +1,9 @@
+/* globals QUnit */
+
 import Ember from 'ember';
 import Ceibo from 'ceibo';
 
-export function throwBetterError(node, key, selector) {
+export function throwBetterError(node, key, selector, stopOnError = false) {
   let path = [key];
   let current;
 
@@ -17,5 +19,14 @@ PageObject: '${path.join('.')}'
   Selector: '${selector}'
 `;
 
-  throw new Ember.Error(msg);
+  let error = new Ember.Error(msg);
+
+  if (stopOnError) {
+    console.error(error);
+    QUnit.stop();
+    alert('Test have been stopped');
+  } else {
+    throw error;
+  }
+
 }
--- a/addon/-private/execution_context/acceptance.js
+++ b/addon/-private/execution_context/acceptance.js
@@ -1,3 +1,4 @@
+/* globals QUnit */
 import {
   guardMultiple,
   buildSelector
@@ -46,7 +47,7 @@ AcceptanceExecutionContext.prototype = {
     let result = find(selector, options.testContainer);
 
     if (result.length === 0) {
-      throwBetterError(this.pageObjectNode, options.pageObjectKey, selector);
+      throwBetterError(this.pageObjectNode, options.pageObjectKey, selector, QUnit.urlParams.stopOnError);
     }
   },
 
@@ -72,7 +73,7 @@ AcceptanceExecutionContext.prototype = {
     result = find(selector, options.testContainer);
 
     if (result.length === 0) {
-      throwBetterError(this.pageObjectNode, options.pageObjectKey, selector);
+      throwBetterError(this.pageObjectNode, options.pageObjectKey, selector, QUnit.urlParams.stopOnError);
     }
 
     guardMultiple(result, selector, options.multiple);
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -22,7 +22,5 @@ module.exports = function(defaults) {
     app.import(compilerPath);
   }
 
-  app.import('vendor/ember-cli-page-object/qunit-configuration.js');
-
   return app.toTree();
 };
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -22,5 +22,7 @@ module.exports = function(defaults) {
     app.import(compilerPath);
   }
 
+  app.import('vendor/ember-cli-page-object/qunit-configuration.js');
+
   return app.toTree();
 };
--- a/index.js
+++ b/index.js
@@ -25,5 +25,10 @@ module.exports = {
 
   _shouldIncludeFiles: function() {
     return this.app.env !== 'production';
+  },
+
+  included: function(app) {
+    this._super.included.apply(this, arguments);
+    app.import('vendor/ember-cli-page-object/qunit-configuration.js', { type: 'test' });
   }
 };
--- a/vendor/ember-cli-page-object/qunit-configuration.js
+++ b/vendor/ember-cli-page-object/qunit-configuration.js
@@ -1,5 +1,11 @@
-/* globals $,QUnit */
+/* globals $,QUnit,Ember*/
 
 $(document).ready(function() {
-  QUnit.config.urlConfig.push({ id: 'stopOnError', label: 'Page object stop on error'});
+  QUnit.config.urlConfig.push({ id: 'stopOnError', label: 'Page Object stop on error'});
+  if (QUnit.urlParams.stopOnError) {
+    Ember.onerror = function () {
+      QUnit.stop();
+      return false;
+    };
+  }
 });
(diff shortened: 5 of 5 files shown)


san650:
Hi @fedekau, I'm really glad that you're working on this feature 🎉 🎉 🍺 🍺 👏 

So, the basic idea of your implementation is correct, we need to register a script to add the checkbox to QUnit toolbar (task 1) and then do something when the checkbox is checked (task 2).

__For the first task__, registering the checkbox to QUnit toolbar, we need to register the javascript that's inside the vendor tree into the end user application, in order to do that we need to use the `included` hook of the index.js file.

Remember that ember-cli-build.js file in addons is used only to setup the dummy test application of the addon, so registering the JS file there only allows the library to run on this addon's tests (which will be needed to test the functionality but nothing more).

You can see [here][1] how to register a js file using the included hook.

__For the second task__, there might be multiple ways of achieving it, what I had in mind when I wrote the issue was to call the `stop` or `pause` helper when the `findWithAssert` helper was unable to find the element. You can see [here][2] where we are validating this (and where we could stop the test).

I didn't imagine something so generic like stoping when any unhandled error occurred on the application but it's an approach we can explore and see how it works out.

Lastly, I imagine that we're going to need to check the state of the checkbox to turn on/off the feature.

There are other minor things to work out after these two tasks but we can tackle them later (like enabling this feature only if the user is using QUnit, avoiding shipping this new js file in production builds, etc.)

[1]: https://ember-cli.com/extending/#importing-dependency-files
[2]: https://github.com/san650/ember-cli-page-object/blob/master/addon/-private/execution_context/acceptance.js#L75

fedekau:
@san650 I have tried removing the import from the `ember-cli-build.js` and adding:

```js
  included: function(app) {
    this._super.included(app);
    app.import('vendor/ember-cli-page-object/qunit-configuration.js', { type: 'test' });
  }
```

to the `index.js` file but all the tests fail, may be I am missing something...

san650:
@fedekau that's strange, can you copy and paste the exact error?

fedekau:
@san650 Here it is:

```
not ok 603 PhantomJS 2.1 - Unit | Property | .create: generates a default scope
    ---
        actual: >
            null
        stack: >
            http://localhost:7357/assets/tests.js:4486:49
            runTest@http://localhost:7357/assets/test-support.js:2791:32
            run@http://localhost:7357/assets/test-support.js:2776:11
            http://localhost:7357/assets/test-support.js:2918:14
            process@http://localhost:7357/assets/test-support.js:2577:24
            begin@http://localhost:7357/assets/test-support.js:2559:9
            http://localhost:7357/assets/test-support.js:2619:9
        message: >
            Died on test #1 http://localhost:7357/assets/tests.js:4483:19
            exports@http://localhost:7357/assets/vendor.js:132:37
            requireModule@http://localhost:7357/assets/vendor.js:32:25
            require@http://localhost:7357/assets/test-loader.js:67:16
            loadModules@http://localhost:7357/assets/test-loader.js:58:25
            load@http://localhost:7357/assets/test-loader.js:89:35
            http://localhost:7357/assets/test-support.js:6472:20: undefined is not a constructor (evaluating '(0, _dummyTestsPageObject.create)({})')
        Log: |
    ...
ok 604 PhantomJS 2.1 - JSCS - unit: unit/create-test.js should pass jscs
ok 605 PhantomJS 2.1 - JSHint | unit/create-test.js: should pass jshint
ok 606 PhantomJS 2.1 - JSCS - unit: unit/shared.js should pass jscs
ok 607 PhantomJS 2.1 - JSHint | unit/shared.js: should pass jshint
ok 608 PhantomJS 2.1 - JSCS - unit: unit/test-helper.js should pass jscs
ok 609 PhantomJS 2.1 - JSHint | unit/test-helper.js: should pass jshint

1..609
# tests 609
# pass  146
# skip  0
# fail  463
Not all tests passed.
Error: Not all tests passed.
    at EventEmitter.App.getExitCode (/Users/federico/fedekau/ember-cli-page-object/node_modules/testem/lib/app.js:402:15)
    at EventEmitter.App.exit (/Users/federico/fedekau/ember-cli-page-object/node_modules/testem/lib/app.js:163:23)
    at /Users/federico/fedekau/ember-cli-page-object/node_modules/testem/lib/app.js:157:14
    at tryCatcher (/Users/federico/fedekau/ember-cli-page-object/node_modules/testem/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/Users/federico/fedekau/ember-cli-page-object/node_modules/testem/node_modules/bluebird/js/release/promise.js:504:31)
    at Promise._settlePromise (/Users/federico/fedekau/ember-cli-page-object/node_modules/testem/node_modules/bluebird/js/release/promise.js:561:18)
    at Promise._settlePromise0 (/Users/federico/fedekau/ember-cli-page-object/node_modules/testem/node_modules/bluebird/js/release/promise.js:606:10)
    at Promise._settlePromises (/Users/federico/fedekau/ember-cli-page-object/node_modules/testem/node_modules/bluebird/js/release/promise.js:685:18)
    at Async._drainQueue (/Users/federico/fedekau/ember-cli-page-object/node_modules/testem/node_modules/bluebird/js/release/async.js:138:16)
    at Async._drainQueues (/Users/federico/fedekau/ember-cli-page-object/node_modules/testem/node_modules/bluebird/js/release/async.js:148:10)
    at Immediate.Async.drainQueues [as _onImmediate] (/Users/federico/fedekau/ember-cli-page-object/node_modules/testem/node_modules/bluebird/js/release/async.js:17:14)
    at processImmediate [as _immediateCallback] (timers.js:383:17)
```

san650:
Push the latest commit so I can take a look at it. If you remove all the `import` statements, from `index.js` and from `ember-cli-build.js` is still failing?

Also, you should use `this._super.included.apply(this, arguments);` to call super.

```js
included: function(app) {
  this._super.included.apply(this, arguments);
  app.import('vendor/ember-cli-page-object/qunit-configuration.js', { type: 'test' });
},
```

fedekau:
@san650 If I put the `QUnit.stop()` before [this line](https://github.com/san650/ember-cli-page-object/blob/master/addon/-private/better-errors.js#L20) it works but the test stops before the exeption is thrown, but I can't put it after because it is unreachable. 

In my opinion throwing the error is very useful for the user, what do you think is a good aproach to solve this? Printing the error to the console with a `console.log`, having some kind of custom event, or just using the `window.onerror` event but filter the errors ?

san650:
@fedekau We don't need to throw the error, we need to write it to the console and then pause the test. We can use `console.error(...)`. Also, I imagine we need to add a banner to explain the user that the test was paused because an error was being thrown and then print the error.

I think the best places to stop the test are:
* https://github.com/san650/ember-cli-page-object/blob/5e5b2c268268401f650e6f14fec7aca3f12f65e4/addon/-private/execution_context/acceptance.js#L75
* https://github.com/san650/ember-cli-page-object/blob/5e5b2c268268401f650e6f14fec7aca3f12f65e4/addon/-private/execution_context/acceptance.js#L49

Mostly because this way we ensure that we're stopping acceptance tests only.

What do you think?

fedekau:
@san650 I have added an alert to test the concept, [here](https://github.com/san650/ember-cli-page-object/pull/234/commits/109ef7573c49e712049f5694c8254c73312fecdf#diff-79f2ca97d3759609d72a960c0d2deef9R27). I will think of something better.

Also I added an optional parameter to the `throwBetterErrors` in order not to have a try/catch outside where it is being called. Does is make sense?

san650:
I'll try the feature on a real project to see how the feature feels, and see if it's as useful as I thought 😛 

I'll keep you posted!

fedekau:
@san650 Did you have time to test this? Was it somewhat helpful?

san650:
Let's make this feature as a separate ember addon that plugs to ember-cli-page-object. @fedekau let's chat about this.

PR #235, opened 2016-08-17
Add page-object helper blueprint

### LINKED ISSUE #173: Unknown blueprint: page-object-helper
Hello.
When I run in terminal:
`ember g page-object-helper login
`

I got an answer:
`Unknown blueprint: page-object-helper`

In file https://github.com/san650/ember-cli-page-object/blob/master/app/blueprints/page-object-helper.js#L1 imported not existing module.

san650:
Fixes #173 

It generates a helper file in `tests/pages/helpers/foo-bar.js` with an example like this

```js
import { findElement } from 'ember-cli-page-object/extend';

export default function fooBar(selector, options = {}) {
  return {
    isDescriptor: true,

    get() {
      return findElement(this, selector, options).is(':disabled');
    }
  };
}
```


san650:
@jeradg @juanazam ready to review

PR #242, opened 2016-08-22
Update components.md

Keeo:
Note that you can define them outside


san650:
Thanks for contributing!

PR #263, opened 2016-11-08
add 'foreach' to arrayDelegateMethods

renpinto:
Added 'forEach' method to arrayDelegateMethods so that it can be used in Collection. Also added a test for this method.


san650:
Thanks for contributing!

Issue #270, opened 2016-12-08
Integration test example misleading


simonihmig:
I find the example of using ember-cli-page-object with integration tests a bit misleading as it is using a page object rather than a page object component. In the given login page example the page object and the component might match, as the page seems to consist only of that login form. But after using this addon a lot for quite a while, I never had this case. Usually the page contains so much more, with nested components/collections, that it does not fit the structure of a component to be tested in isolation. 

So usually when creating a new component, I also create a new matching page object component, that could be added to page object for acceptance tests, but for the component integration test they look something like this one:

```js
import { expect } from 'chai';
import { describeComponent, it } from 'ember-mocha';
import {
  beforeEach,
  afterEach
} from 'mocha';
import hbs from 'htmlbars-inline-precompile';
import navigationToggle from 'frontend/tests/pages/components/navigation-toggle';
import {
  create
} from 'ember-cli-page-object';

// Create a page object containing only our page object component
const page = create({
  navigationToggle
});
const component = page.navigationToggle;

describeComponent('navigation-toggle', 'Integration | Component | navigation toggle',
  {
    integration: true
  },
  function() {
    beforeEach(function() {
      page.setContext(this);
    });

    afterEach(function() {
      page.removeContext();
    });

    it('shows icon', function() {
      this.render(hbs`{{navigation-toggle icon="menu"}}`);
      expect(component.hasMenuIcon).to.be.true;
    });
  }
);
```

It seems you cannot use the page object component directly without it being part of a page object, so that's why I have this page setup before the tests.

Is this how it should be done? If so, would you consider changing the docs accordingly? Can also create a PR if you want...

san650:
Hi @simonihmig, sorry for the late response on this issue.

I see what you're saying and I think the problem goes beyond the example and has to do with the fact that we use the name "page object component". That name is misleading because it makes you expect that this "page object component" will behave like a page object when in fact it's just a plain old JavaScript object without any special behavior.

When you plug one of these "page object components" into a page object, they gain the special behavior we expect to see in a page object.

So to sum up, in ember-cli-page-object, a "page object component" is just a plain old JavaScript object and if we want to create a page object from it, we need to call the `create` function on it.

Let me share with you an example of how I'm using it:

`tests/pages/components/login-form.js`

```js
import {
  clickable,
  create,
  fillable,
  text,
  visitable
} from 'ember-cli-page-object';

const definition = {
  username: fillable('#username'),
  password: fillable('#password'),
[...]

PR #274, opened 2016-12-28
[WIP] Component/-test blueprint with page-object integration

simonihmig:
So after the xmas break I was able to spend some time on this, so here we go... :)

This is a work in progress PR for the blueprint feature discussed in #270. Should be fully functional, manually tested with ember-cli-qunit, ember-cli-mocha 0.11 and 0.13. node-tests for the new blueprints are missing though. Can add some later when there are no concerns regarding the feature itself...

Most of the code is copied from the original blueprint definitions ([ember-cli-legacy-blueprints](https://github.com/ember-cli/ember-cli-legacy-blueprints)), with the customizations added on top. See the separate commits.

### ToDo 

- [x] add tests
- [x] fix when used in pod-structured apps. See https://github.com/ember-cli/ember-cli/issues/6637


simonihmig:
After sleeping a night over it, just realized that overriding the component blueprint was totally useless. Removed it, now it only overrides component-test. So less maintenance...

san650:
Thanks @simonihmig I'll take a look at it today.

simonihmig:
Hi @san650, did you have a chance to look at it? 

Just fixed one issue, should hopefully work fine now. If you have no objections, I would try to add some tests then...

san650:
@simonihmig I didn't have much time to look at it. My first impression is that it's a lot of code to maintain in sync with official's blueprint.

Just thinking out loud, wouldn't it better to have a separate addon for this? My main concern is that if we break an official blueprint then people might end up uninstalling this addon because of this.

What about creating a `ember-cli-page-object-blueprints` addon? And in time, if everything works well, we might merge it back to ember-cli-page-object.

simonihmig:
Well, would be possible, and I am not heavily opposed to this. But I am afraid that users might miss out the benefits of this when this is just an *optional* "addon-addon" to the main addon. Maybe we could include the addon-addon by default using an [addAddonToProject](https://ember-cli.com/api/classes/Blueprint.html#method_addAddonToProject) call in the default blueprint, and users could then opt-out if needed? (just be removing this from their `package.json`)

But maybe let's look at your concerns first:
* only the component-test blueprint is overridden, anything else should be untouched
* this blueprint uses just public APIs of ember-cli (with one exception, but this seems to be declared public soon, see https://github.com/ember-cli/ember-cli/pull/6645). So I don't see a reason why it should stop working out of nowhere, especially when tests have been added.
* the only thing that has been literally copy-pasted and which users might expect to be in sync, is the original component-test template (https://github.com/san650/ember-cli-page-object/pull/274/files#diff-178df03461b3c274c1b2610d7ca1814e). When that changes in ember-cli-legacy-blueprints, one would have to sync it back. But that is just a matter of copy and paste again, nothing in that file has been modified. And that was basically the idea why we decided to override it and not try to access private APIs to reuse the original blueprint.
* the only scenario I see that might make real trouble is if there is some real functionality added to the original blueprint (the `index.js` file, not the template), like for example support for a new file layout (like the [module unification RFC](https://github.com/emberjs/rfcs/blob/master/text/0143-module-unification.md)) and we are not able to catch up in a timely manner

So, you have certainly the last word on this! ;)

san650:
@simonihmig You're right, let's move on with this PR, let's add some tests. Thanks for the hard work!

simonihmig:
Happy to do so!

The original blueprint tests for `component-test` are actually quite verbose, testing all the different scenarios (app/addon, pod, nested, custom path etc.): https://github.com/ember-cli/ember-cli-legacy-blueprints/blob/master/node-tests/blueprints/component-test.js

Not sure if we should maintain the same verbosity level here? If so could basically copy/paste all of these for the default (unmodified) blueprint and add another set for the custom version with the `--page-object` flag applied. What do you think?

san650:
Just having some tests for a sanity check is enough. One with the `--page-object` flag and another one without it.

simonihmig:
@san650 added the tests. Ended up with a couple more tests, mainly to cover the different test frameworks. But was again a lot of copy paste (getting really good at that! 😎) from the original ones, so no big deal. Should now cover most scenarios.

From my point of view this should be ready to merge!

simonihmig:
While the `node-tests` run fine on my local machine, it seems they are not configured for Travis! Aware of that?

san650:
Thanks for working on this, I'll take a look at the PR and merge it.

> While the node-tests run fine on my local machine, it seems they are not configured for Travis! Aware of that?

I thought they were, thanks for the heads up. Would you mind updating the `.travis.yml` to also run node tests? You can open a separate PR if you want.

simonihmig:
Actually adding some documentation would also be required. Can add something to https://github.com/san650/ember-cli-page-object#blueprints, but not sure about the "real" documentation site? Will you eventually add that?

> I thought they were, thanks for the heads up. Would you mind updating the .travis.yml to also run node tests? You can open a separate PR if you want.

Ok, I can do that.

san650:
@simonihmig please add the documentation to the README. We don't have any documentation about the blueprints in the current website.

We're working on a new website which will have a lot more info, I'll add the info about the blueprints to it.

If you are curios, here's the guide about blueprints for the new website (WIP) https://github.com/san650/ember-cli-page-object/blob/site-redesign/guides/blueprints.md If you want to also create a PR for it, it would be great 😃

simonihmig:
Updated the README, hope that this is ok!

Will probably add a PR to the website some time later. Feel free to ping me if I forget this! 😉

simonihmig:
Don't know why Travis failed, but does not seem to be related to my changes. @san650 can you rerun?

Anything else missing?

san650:
@simonihmig Didn't have the time to review it. I'll try to do it tomorrow. I've restarted the build

simonihmig:
No hurry! Thanks for rerunning build, seems I forgot to save some deps to `package.json`, should be fixed now...

san650:
Thank you so much for working on this!

simonihmig:
@san650 You are welcome! Thanks for cutting out a release so quickly, I can now remove the fork from our projects, was already using it here and there! :)

> You can create a new page object called users using the generate command

Btw, this is in your release notes, probably by mistake! https://github.com/san650/ember-cli-page-object/releases

san650:
@simonihmig copy and paste error 😛 

Fixed it. Thanks!

PR #275, opened 2017-01-11
add node-tests to CI

simonihmig:
This should run `npm run nodetest` on Travis as a separate parallel job.


san650:
Thanks for contributing!

PR #285, opened 2017-01-30
Fix jsdoc lining errors

hidnasio:
The linter can be run like `find addon -type f -name '*.js' -exec node node_modules/documentation/bin/documentation.js --lint --shallow {} \;`


san650:
Thanks for contributing!

PR #286, opened 2017-02-06
Add appendix about composability to quickstart

sukima:


cs3b:
I like the way you show flexibility of the tool. I've also written article how I nest Page Object within Page Object :-) https://medium.com/@cs3b/acceptance-testing-with-page-object-nesting-8c3c7384f367

When I was writing the article, It remind me that if we would have feature like `with` similar to `as` recently added that would change only scope (not the object, as the `as` is doing) Then it would be even easier to work with page objects. (Just my thoughts :-)

san650:
@sukima sorry for the long delay merging this, it looks really nice. Thanks for contributing!

san650:
@cs3b Thanks for sharing the post, it's very interesting to see how people solves this problem with different approaches.

About `with` helper, I think we don't have any feature request with that, It could be a nice addition although I don't know how we could implement it. Please, feel free to open the feature request if you think its worth it :)

PR #290, opened 2017-03-29
Add value as default component attribute

backspace:
This is true according to the 1.6.0 release notes:
https://github.com/san650/ember-cli-page-object/releases/tag/v1.6.0

Note that these don’t exactly align to the links in each file as I believe they are incorrect… another PR forthcoming.


san650:
Thanks for fixing this!

PR #291, opened 2017-03-29
Correct property links to be in current version

backspace:
Hey, it seemed to me that these links should be within the version you’re currently looking at, rather than always to 1.6.x, do you agree?


backspace:
I see this has conflicts now that you merged my other PRs, I’ll rebase hehe

san650:
Thanks!

PR #292, opened 2017-03-29
Fix typo: explicitely to explicitly

backspace:
Thanks for all your work on this, I am still finding it so useful! Hopefully this triple-PR situation isn’t annoying; if it’s preferable, I could make them all one!


san650:
It's fine either way, if you prefer separate PRs, please continue to do so.

Thanks for contributing!

Issue #295, opened 2017-04-04
Enable more concise configuration of top-level properties that abstract component details


magistrula:
I really like organizing page objects in terms of components because I think doing so helps to make page objects coherent and well-organized, but it can lead to tests that are tightly coupled to the page structure.

Instead of writing tests that look like this:
```javascript
it('disables the Roster form if the user inputs the name "Voldemort"', function() {
  RosterPageObject.form.input.fillIn('Voldemort');
  RosterPageObject.form.submitButton.click();

  expect(RosterPageObject.form.input.isDisabled).to.be.true;
});
```
I'd like to write tests that look like this:
```javascript
it('disables the Roster form if the user inputs the name "Voldemort"', function() {
  RosterPageObject.addName('Voldemort');

  expect(RosterPageObject.isNameInputDisabled).to.be.true;
});
```
Currently, this is how I'd set up my page objects if I want to write tests as shown in the second snippet above:
```javascript
// Imagine that MyForm is a generic form component, 
// actually defined in its own page object file.
const MyForm = {
  scope: '.my-form',

  input: {
    scope: 'input',
    isDisabled: is(':disabled')
  },

  submitButton: {
    scope: 'button'
  }
};

// This is the page object for the component we're going to test
const RosterPageObject = create({
  scope: '.roster',

  form: MyForm,
  
  addName(name) {
    this.form.input.fillIn(name);
    this.form.submitButton.click();
  },

  isNameInputDisabled: {
    isDescriptor: true,
    get() {
      return this.form.input.isDisabled;
    }
  }
});
```
It would be nice to have some helpers (e.g., `component`, `pipe` and `alias` in the examples below) that make setting up these top-level properties easier and more concise.
```javascript
const MyForm = {
  scope: '.my-form',

  input: component('input', {
    isDisabled: is(':disabled')
  },

  submitButton: component('button')
};

const RosterPageObject = create({
  scope: '.roster',

  form: MyForm,
  
  addName: pipe('form.input.fillIn', 'form.submitButton.click'),
  isNameInputDisabled: alias('form.input.isDisabled')
});
```
The helpers I have in mind so far are: 
- `alias`: sets up a property that returns the value of some other property
- `component`: sugar for setting up a POJO with a `scope` property
- `conditional-alias`: sets up a property that can return various different values depending on some set of conditions to be checked each time this property is accessed
- `ternary-alias`: a simpler version of `conditional-alias`, with only one condition to check
[...]

PR #296, opened 2017-04-07
Add alias helper

magistrula:
### Purpose
- Add an `alias` helper that can be used to define PageObject properties that are aliases of other properties. This is especially useful when seeking to expose something like `form.submitButton.click` as `submitForm`. 
- Addresses https://github.com/san650/ember-cli-page-object/issues/295

### Test Coverage
- [x] `objectHasProperty` & `getProperty` are unit tested in `utils-test`
- [x] `alias` helper tested against acceptance & integration adapters in `properties/alias-test`
- [x] Chaining of aliased actions tested in `acceptance/actions-test`
- [x] Chaining of aliased actions tested in `integration/actions-test`
- [x] Alias export verified in `unit/addon-exports-test`

### Open Questions
- Are there any style inconsistencies between this code and the existing code that need to be resolved?


addon/-private/properties/alias.js
```diff
@@ -0,0 +1,64 @@
...
+ * });
+ *
+ * // checks value of `page.continueButton.text`
+ * assert.equal(page.continueText, 'Go ahead!');
+ *
+ * @example
+ *
+ * const page = PageObject.create({
+ *   clickContinueButton: clickable('button.continue'),
+ *   continue: alias('clickContinueButton')
```
magistrula: I don't know why anyone would create an alias in this situation; seems like you would just name `clickContinueButton` as `continue` instead. Is it even worth including this example?

addon/-private/properties/alias.js
```diff
@@ -0,0 +1,64 @@
...
+ * });
+ *
+ * // calls `clickContinueButton`
+ * page.continue();
+ *
+ * @public
+ *
+ * @param {string} pathToProp - dot-separated path to PageObject property
```
san650: I think this description is very important to remark in the documentation.

What will be __supported__ by the add-on is the ability to point to other PageObject's property and not to any property.

```js
const myDescriptor = {
  isDescriptor: true,
  get(key) {
    return {
      hello: 'world'
    };
  }
};

  foo: myDescriptor,

  aliased: alias('foo.hello')
});
```

Although this works, we should consider this as a _not supported_ scenario -- i.e. we don't promise this is going to continue working on a future version of the addon -- because `hello` is not a property of a page object, it's a property of a generic object returned by a page object's property.

san650: There are some differences between `Ember.computed.alias` and `PageObject.alias` that IMHO we should highlight in the documentation so people doesn't get frustrated by things not working as they expect. One I can think of is how we treat collections, with Ember's alias computed property you can do things like:

```js
import Ember from 'ember';

export default Ember.Component.extend({
  posts: Ember.computed(function() {
    return ['hello', 'world', 'foo', 'bar'];
  }),

  firstPost: Ember.computed.alias('posts.firstObject')
});
```

We aren't supporting these _operators_ (`firstObject`, `lastObject`) and I think having a note in the docs mentioning that could save us from some bug reports.

What do you think?

addon/index.js
```diff
@@ -23,6 +24,7 @@ export { findElementWithAssert } from './-private/extend/find-element-with-asser
 export { buildSelector, getContext } from './-private/helpers';
 
 export default {
+  alias,
```
san650: Can you add `alias` to the exports test? https://github.com/san650/ember-cli-page-object/blob/217d647ec34ad8b1686d824d16a2253fbcd5e22b/tests/unit/addon-exports-test.js

san650:
I forgot to mention that we have an AUTHORS file, can you add your name, email and github handler to the bottom of the file?

https://github.com/san650/ember-cli-page-object/blob/master/AUTHORS

tests/acceptance/actions-test.js
```diff
@@ -48,6 +54,20 @@ test('allows to chain actions', function(assert) {
   });
 });
 
+test('allows to chain aliased actions', function(assert) {
```
magistrula: @san650, here is the failing test; I had neglected to push it.

san650: @magistrula now I see what the problem is:

When you invoke an action, the action returns the current page object context so you can chain other actions. The problem here arises when nesting properties:

```js
const page = create({
  foo: {
  },

  bar: clickable('btn')
});

page.foo.clickOn('lorem ipsum').bar(); // fails
```

The call to `clickOn` returns the object `foo` and not the root page object, so `bar` is not found. This is the default behavior of the page object and it has been working fine so far.

The question is, how should `alias` behave when you use a nested action? should it return the aliased action context or the alias context?

```js
const page = create({
  foo: {
  },

  bar: clickable('btn'),

  fooClickAlias: alias('foo.clickOn')
});

page.fooClickAlias('lorem ipsum').bar(); // fails
```

My guess would be that it should return whatever the action returns (because is just a function after all).

So, updating the acceptance test to something like this should be enough

```diff
--- a/tests/acceptance/actions-test.js
+++ b/tests/acceptance/actions-test.js
@@ -18,16 +18,16 @@ let page = PageObject.create({
   keys: {
     clickOn: clickOnText('.numbers'),
     sum: clickable('button', { scope: '.operators', at: 0 }),
-    equal: clickable('button', { scope: '.operators', at: 2 })
+    equal: clickable('button', { scope: '.operators', at: 2 }),
+    add: alias('sum'),
+    getResult: alias('equal')
   },

   screen: value('.screen input'),
   fillValue: fillable('.screen input'),

   visitCalculator: alias('visit'),
-  clickOnKey: alias('keys.clickOn'),
-  add: alias('keys.sum'),
-  getResult: alias('keys.equal')
-  getResult: alias('keys.equal')
+  clickOnKey: alias('keys.clickOn')
 });

 test('allows to chain actions', function(assert) {
 });

 test('allows to chain actions', function(assert) {
@@ -58,13 +58,12 @@ test('allows to chain aliased actions', function(assert) {
   page
     .visitCalculator()
     .clickOnKey('1')
-    .clickOnKey('2')
-    .add()
-    .clickOnKey('3')
+    .sum()
+    .clickOn('2')
     .getResult();

   andThen(function() {
-    assert.equal(page.screen, '15');
+    assert.equal(page.screen, '3');
   });
 });
```

---

To continue with the other discussion, if `alias` should change or not the return value of the actions, we could add an option to `alias` in the future if we see the need for it, something like

```js
const page = create({
  foo: {
  },

  bar: clickable('btn'),

  fooClickAlias: alias('foo.clickOn', { returnSelf: true })
});

page.fooClickAlias('lorem ipsum').bar(); // works
```

What do you think?

magistrula: Aha...this makes complete sense. I feel like it would be most intuitive to for aliased methods to return the alias context. Given a page object that looks like this...
```javascript
const page = create({
  foo: {
  },

  bar: clickable('.btn1'),
  baz: clickable('.btn2'),

  clickFoo: alias('foo.click'),
  fillFoo: alias('foo.fillIn')
});
```
I think it would be unexpected if you could do `page.bar().baz()` but not `page.clickFoo().fillFoo('stuff')`. Does that seem like a reasonable point in favor of returning the alias context rather than the aliased action context?

addon/-private/properties/alias.js
magistrula: Great point. Would you suggest explicitly calling this out as an unsupported use case in the documentation? Even though it's "unsupported," should I add a test for this scenario? That way, if at some point this no longer works, a test failure will sound the alarm and then a decision can be made about whether this use case should continue to be supported or whether the test should simply be removed and the documentation updated accordingly.

magistrula: Definitely an important point that `PageObject.alias` should be conceptually consistent with `Ember.computed.alias`. It seems like the `firstObject`/`lastObject` issue is mainly a difference between `PageObject.collection` and `Ember.Array` rather than a difference between `PageObject.alias` and `Ember.computed.alias`. Does that seem accurate?

If we have aliased methods return the alias context instead of the aliased method context, the helper name `alias` would be inaccurate, because in that case it's more than an alias...it's a transfer of ownership from one context to another. Hmmmmm. Maybe I need to rethink the name `alias`?

addon/-private/properties/alias.js
```diff
@@ -0,0 +1,67 @@
...
+
+      if (typeof value !== 'function') {
+        return value;
+      }
+
+      const executionContext = getExecutionContext(this);
+
+      return function(...args) {
+        return executionContext.runAsync(() => value(...args));
+      };
```
magistrula: This is my proposed solution for having aliased methods return the alias context.

tests/acceptance/actions-test.js
san650: While this fixes the problem for actions, I think it breaks the behavior for aliasing regular functions or collections where the return value must remain unchanged.

I would suggest to keep `alias` simple and just return the value of the original function and have an option to change the behavior if desired. So if the user wants to return the alias context, then the user should make it explicit in the alias definition `fooClick: alias('foo.click', { returnAliasContext: true })` (or some other name for the option).

We could go even further and have separated helpers: `alias` and `actionAlias`.

magistrula: Do you think `{ chainable: true }` would be a reasonable option to pass in, or is it better go with something more descriptive of what is actually happening (i.e., `{ returnAliasContext: true }`).

addon/-private/properties/alias.js
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * @throws Will throw an error if the PageObject does not have the specified property.
+ */
+export function alias(pathToProp, options = {}) {
+  return {
+    isDescriptor: true,
+
+    get(key) {
+      if (!objectHasProperty(this, pathToProp)) {
+        throw new Ember.Error(
```
magistrula: Hrm...I think I can make this a bit more consistent with the error message that `throwBetterErrors` generates.

magistrula: The easiest thing would be to merge #301 first so that I can build on those changes to `throwBetterErrors` here. Or, if this PR is ok to merge first, I can follow up with an improvement to the error messaging right after #301 merges.

san650: We can improve this later, let's merge it as it is.

san650:
@magistrula I really like the implementation, I know it's not easy to understand the internals of the addon and you have mastered it 😃 .

Sorry the long time it took me to review this. I'll merge it now.

Issue #294, opened 2017-04-04
Fillable support for `contenteditable` elements


robustdj:
If I have a template like:

```html
<div contenteditable class="hour">11</div>
```

It would be nice to use the `fillable` helper to change the value.

san650:
Implemented in #301

PR #301, opened 2017-04-12
Add fillable support for contenteditable elements

magistrula:
### Purpose
- Configure `fillable` to handle both inputs and contenteditable elements
- Addresses https://github.com/san650/ember-cli-page-object/issues/294

### Test Coverage
- `fillable` behavior tested against inputs & contenteditables in `acceptance/actions-test` and `integration/actions-test`
- `fillable-test` verifies the following:
  - If a `clue` is passed to fillable, `contenteditable` elements are included in the search for an element to fill.
  - If the element targeted by `fillable` has `contenteditable="false"`, an error is thrown.
- `value-test` verifies that the helper returns the html content of contenteditable elements.

### TODOS
- [x] Update `fillable` & `value` documentation with contenteditable examples
- [x] Update `value` helper to return html of contenteditable element

### Open Questions 
- [x] The ember async helper [fillIn](https://github.com/emberjs/ember.js/blob/master/packages/ember-testing/lib/helpers/fill_in.js) does not support contenteditable elements. Is this a reasonable workaround for now?
- [x] An element like `<div contenteditable</div>` is editable even though the `contenteditable` attribute is not given a value. While it's probably not the best practice to leave out a value, it's probably not up to the `fillable` helper to enforce that, so the changes in this PR treat the following as valid contenteditable targets: `[contentditable]`, `[contenteditable=""]`, `contenteditable="true"`, `contenteditable="anythingbutfalse".` Does that seem appropriate?


magistrula:
@san650, I still have some TODOS, but what do you think of the general approach?

addon/-private/properties/fillable.js
```diff
@@ -1,8 +1,26 @@
...
+
+export function fillElement($el, content, { selector, pageObjectNode, pageObjectKey }) {
+  if ($el.is(EDITABLE_CONTENTEDITABLE)) {
+    $el.html(content);
+  } else if ($el.is(UNEDITABLE_CONTENTEDITABLE)) {
+    throwBetterError(pageObjectNode, pageObjectKey, selector, CONTENTEDITABLE_IS_FALSE);
+  } else {
+    $el.val(content);
+  }
+}
```
san650: Can we move this helper (`fillElement`) to a file named `addon/-private/execution_context/helpers.js`? I don't want for execution contexts to know about page object properties and this function is used only on the execution contexts.

san650:
@magistrula I like the implementation and nice catch about `value` helper!

> The ember async helper fillIn does not support contenteditable elements. Is this a reasonable workaround for now?
I think so, I was reading ember's implementation of `fillIn` and seems close enough. I know they are using real DOM events instead of jQuery but we're relying on jQuery for now.

>  An element like <div contenteditable></div> is editable even though the contenteditable attribute is not given a value. While it's probably not the best practice to leave out a value, it's probably not up to the fillable helper to enforce that, so the changes in this PR treat the following as valid contenteditable targets: [contentditable], [contenteditable=""], contenteditable="true", contenteditable="anythingbutfalse". Does that seem appropriate? 

Yes, it seems appropriate.

Thanks for working on this and let me know if I can be of any help.

addon/-private/helpers.js
```diff
@@ -309,3 +313,37 @@ export function getProperty(object, pathToProp) {
...
+ * @param {string} options.selector       Element selector
+ * @param {Ceibo} options.pageObjectNode  PageObject node containing the method
+ * which, when invoked, resulted in this call to `fillElement`
+ * @param {string} options.pageObjectKey  Key of method on PageObject
+ * which, when invoked, resulted in this call to `fillElement`
+ * @return
+ *
+ * @throws Will throw an error if called on a contenteditable element that has
+ * `contenteditable="false"`
+ */
```
magistrula: Is this level of documentation right for this kind of helper? Is it correct to mark it `@public` because it is imported by the execution contexts, even though it is not exposed in the actual add-on exports? Now that this is in `helpers.js`, does it need its own unit test?

san650: I would move it to a new helper file inside the execution context folder so we avoid people using it on any property directly `/addon/-private/execution_context/helper.js`.

The helpers in this file are used all over the place but for this one is better to avoid it's widespread use.

> Is it correct to mark it @public because it is imported by the execution contexts

Mark it as @private please. 

> Is this level of documentation right for this kind of helper? 

It's not that necessary at this level of helpers but it's nice to have.

> does it need its own unit test?

I think that by testing `fillable` and `selectable` is enough to ensure that this function is working properly. On the other hand if there are some scenarios you think are important to cover with unit tests, feel free to add those, but I think we're good as it is right now.

magistrula: Ah, got it … misread your original comment about what file to put this in … will fix.

addon/-private/execution_context/helpers.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @param {jQuery} $selection              jQuery object containing collection of
+ * DOM elements to fill in
+ * @param {string} content                 Content to be inserted into target element(s)
+ * @param {Object} options                 Options for error reporting
+ * @param {string} options.selector        jQuery selector
+ * @param {Ceibo} options.pageObjectNode   PageObject node containing the method
+ * which, when invoked, resulted in this call to `fillElement`
+ * @param {string} options.pageObjectKey   Key of method on PageObject
+ * which, when invoked, resulted in this call to `fillElement`
```
magistrula: What's the proper indentation convention for wrapping lines in a comment like this?

addon/-private/execution_context/acceptance.js
```diff
@@ -33,13 +39,21 @@ AcceptanceExecutionContext.prototype = {
...
 
-  fillIn(selector, container, text) {
-    /* global fillIn */
-    if (container) {
-      fillIn(selector, container, text);
-    } else {
-      fillIn(selector, text);
-    }
+  fillIn(selector, container, options, content) {
+    let $selection = find(selector, container || findClosestValue(this.pageObjectNode, 'testContainer'));
```
magistrula: Note: I changed this from `$element` to `$selection`...seems more accurate, since this could be a collection of multiple elements.

magistrula:
@san650, any thoughts on what is causing these test failures? 
https://travis-ci.org/san650/ember-cli-page-object/jobs/222452343

san650:
@magistrula try rebasing with master, I've fixed some strange errors with travis-ci / ember-data / ember-cli-addon-tests. I think the latest version of ember-data broke some of these acceptance tests

addon/-private/execution_context/helpers.js
san650: The format that work best with the tool that generates the API from the jsdocs is

```js
/**
* @private
*
* Fills inputs, textareas, or contenteditable elements with the passed-in content.
* @param {jQuery} $selection - jQuery object containing collection of * DOM elements to fill in
* @param {string} content - Content to be inserted into target element(s)
* @param {Object} options - Options for error reporting
* @param {string} options.selector - jQuery selector
* @param {Ceibo} options.pageObjectNode - PageObject node containing the method which, when invoked, resulted in this call to `fillElement`
* @param {string} options.pageObjectKey - Key of method on PageObject which, when invoked, resulted in this call to `fillElement`
```

I know that some lines are too long, but let's ignore that for now. I can change the format in the future.

san650:
More info about the issue: https://github.com/san650/ember-cli-page-object/commit/ad07e458f8ceaf3a0b9d53b835e1c87128c064a8

magistrula:
Hmm, I have commit ad07e45 on this branch, but I those build tests failed twice then mysteriously passed.

san650:
@magistrula please, let me know when the PR is ready so I can review it / merge

magistrula:
@san650, the code changes are ready for review, but I'm getting inconsistent failures in these two test scenarios: `generates a development build` and `generates a production build`. Is there anything this PR should do to address that issue?

san650:
@magistrula The inconsistent failures are not related to this PR, I'll merge it.

PR #303, opened 2017-04-19
Use throwBetterError in alias helper

magistrula:
### Purpose
- Improve error-reporting of `alias` helper by using `throwBetterError`

### Approach
- `throwBetterError`: make `selector` part of an optional `options` parameter so that it can be excluded when there is no selector to include in the error message. 
- `alias`: use `throwBetterError` to build the `Ember.Error` thrown when an aliased property is not found on the page object.


juanazam:
@magistrula this (and the alias property as a whole) looks great! 👏

magistrula:
Thanks for the review, @juanazam!

Issue #304, opened 2017-04-20
new Ember.Error in better-error not logging the message in the console


Leooo:
using Ember 2.10.2, on Chrome

https://github.com/san650/ember-cli-page-object/blob/master/addon/-private/better-errors.js#L22

the message is not logged in the console

add a `console.error` call before the error to make sure it is logged?

san650:
What's strange is that errors were being logged before, so something changed (in Ember?) that makes this error not to be logged anymore.

I remember that was a change in how backburner reported exceptions when env was testing, see https://github.com/emberjs/ember.js/pull/14898. Maybe is related to that.

Having the exception in the console is super useful so I think we should log it ourself if it's not printed by Ember.

PR #306, opened 2017-05-02
throwBetterError logs error to console

magistrula:
### Purpose
- Ensure that `throwBetterError` logs an error to the console.
- Addresses #304 

### Approach
- Call `Ember.Logger.error` before throwing `Ember.Error`.

### Test Coverage
- Added unit tests for `throwBetterError`, which I probably should have done in #303, since `throwBetterError` now has some logic around whether or not to report a selector.
- I wasn't sure at first where to put the test for `throwBetterErrors`, so I reshuffled the `tests` folders a bit.

### Open Questions
- [x] Are these changes to the `tests` folder structure helpful/sensible?
- [x] I sort of arbitrarily decided on `Ember.Logger.error` rather than `console.error`. It doesn't make much of a difference either way...any reason to prefer one over the other?

### DIFF
--- a/tests/unit/-private/better-errors-test.js
+++ b/tests/unit/-private/better-errors-test.js
@@ -1,0 +1,60 @@
+import Ember from 'ember';
+import { test, module } from 'qunit';
+import { create } from 'ember-cli-page-object';
+import {
+  throwBetterError
+} from 'ember-cli-page-object/-private/better-errors';
+
+const page = create({
+  foo: {
+    scope: '.foo',
+    bar: {
+      scope: '.bar',
+      focus() {}
+    }
+  }
+});
+
+module('Unit | throwBetterError');
+
+test('shows the expected error message when `selector` is not passed in', function(assert) {
+  assert.expect(1);
+
+  const fn = () => {
+    throwBetterError(page.foo.bar, 'focus', 'Oops!');
+  };
+  const expectedError = new Ember.Error(
+    "Oops!\n\nPageObject: 'page.foo.bar.focus'"
+  );
+
+  assert.throws(fn, expectedError, 'should show message & property path');
+});
+
+test('shows the expected error message when `selector` is passed in', function(assert) {
+  assert.expect(1);
+
+  const fn = () => {
+    throwBetterError(page.foo.bar, 'focus', 'Oops!', { selector: '.foo .bar' });
+  };
+  const expectedError = new Ember.Error(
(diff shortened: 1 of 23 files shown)


tests/unit/-private/better-errors-test.js
```diff
@@ -0,0 +1,57 @@
...
+});
+
+test('logs the error to the console', function(assert) {
+  assert.expect(2);
+
+  const origEmberLoggerError = Ember.Logger.error;
+
+  Ember.Logger.error = (msg) => {
+    assert.equal(msg, "Oops!\n\nPageObject: 'page.foo.bar.focus'");
+    Ember.Logger.error = origEmberLoggerError;
```
san650: we might have to restore this on an `after` hook or a `try/finally` block.

magistrula: Ah, I see...in case the test fails and line 53 is never hit? Are you thinking something like this?
```javascript
Ember.Logger.error = (msg) => {
  try {
    assert.equal(msg, "Oops!\n\nPageObject: 'page.foo.bar.focus'");
  } finally {
    Ember.Logger.error = origEmberLoggerError;
  }
};
```

san650: Yes, I was thinking that if something fails, line 53 could not be executed.

The solution I have in mind is

```js
test('logs the error to the console', function(assert) {
  assert.expect(2);

  const origEmberLoggerError = Ember.Logger.error;

  try {
    Ember.Logger.error = (msg) => {
      assert.equal(msg, "Oops!\n\nPageObject: 'page.foo.bar.focus'");
    };

    assert.throws(() => throwBetterError(page.foo.bar, 'focus', 'Oops!'));
  } finally {
    Ember.Logger.error = origEmberLoggerError;
  }
});
```

So we restore `Ember.Logger.error` even if the method is not called.

san650:
About your open questions

> Are these changes to the tests folder structure helpful/sensible?
I like the idea of replicating the`/addons/` structure inside  in the unit test folder.

> I sort of arbitrarily decided on Ember.Logger.error rather than console.error. It doesn't make much of a difference either way...any reason to prefer one over the other?

Let's go with `Ember.Logger.error` :)

tests/unit/-private/better-errors-test.js
magistrula: Fixed.

magistrula:
Ready for final review, @san650.

san650:
@magistrula looks really nice! Feel free to merge it to master.

PR #316, opened 2017-06-10
Triggerable returns function with eventProperties parameter

magistrula:
### Purpose
Some events can be triggered without any specified properties (e.g., `mousedown`), but others require properties like `which` (e.g., `keydown`). `triggerable` currently requires event properties to be defined at the time when `triggerable` is called in the page object, but it is sometimes more convenient to specify event properties at the time when the triggerable descriptor is used in tests. This PR modifies `triggerable` to accept either static event properties defined at the time when `triggerable` is called in the page object or dynamic event properties passed in when the descriptor is accessed.

### Test Coverage
- New tests verify that `triggerable` descriptor applies the event properties.
- New tests verify that if event properties are specified both in the `triggerable` call and the `triggerable` descriptor call, the props passed into the descriptor are applied.

### Open Questions
- [ ] Is it appropriate for `triggerable` to have two different ways of receiving `eventProperties`? Would it be better to define a separate `dynamicTriggerable` helper?

### DIFF
--- a/tests/unit/-private/properties/triggerable-test.js
+++ b/tests/unit/-private/properties/triggerable-test.js
@@ -22,23 +22,58 @@ moduleForProperty('triggerable', function(test) {
     return this.adapter.wait();
   });
 
-  test("calls Ember's triggerEvent helper with event options", function(assert) {
-    assert.expect(3);
+  test("calls Ember's triggerEvent helper with static event options", function(assert) {
+    assert.expect(1);
 
-    let expectedSelector = 'span';
     let page = create({
-      foo: triggerable('keypress', expectedSelector, { eventProperties: { keyCode: 13 } })
+      foo: triggerable('keypress', 'span', { eventProperties: { keyCode: 13 } })
     });
 
     this.adapter.createTemplate(this, page, '<span></span>');
 
     this.adapter.triggerEvent((actualSelector, _, event, options) => {
-      assert.equal(actualSelector, expectedSelector);
-      assert.equal(event, 'keypress');
       assert.equal(options.keyCode, 13);
     });
 
     page.foo();
+
+    return this.adapter.wait();
+  });
+
+  test("calls Ember's triggerEvent helper with dynamic event options", function(assert) {
+    assert.expect(1);
+
+    let page = create({
+      foo: triggerable('keypress', 'span')
+    });
+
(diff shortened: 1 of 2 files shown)


tests/unit/-private/properties/triggerable-test.js
```diff
@@ -22,19 +22,16 @@ moduleForProperty('triggerable', function(test) {
...
     let page = create({
-      foo: triggerable('keypress', expectedSelector, { eventProperties: { keyCode: 13 } })
+      foo: triggerable('keypress', 'span', { eventProperties: { keyCode: 13 } })
     });
 
     this.adapter.createTemplate(this, page, '<span></span>');
 
     this.adapter.triggerEvent((actualSelector, _, event, options) => {
-      assert.equal(actualSelector, expectedSelector);
-      assert.equal(event, 'keypress');
```
magistrula: These lines can be removed because these expectations are covered in the first test in this file. When I found myself adding two more tests like this one, I thought it would be best to omit redundant assertions to keep the test trim & readable.

tests/unit/-private/properties/triggerable-test.js
```diff
@@ -43,6 +40,44 @@ moduleForProperty('triggerable', function(test) {
...
+    this.adapter.triggerEvent((actualSelector, _, event, options) => {
+      assert.equal(options.keyCode, 13);
+    });
+
+    page.foo({ keyCode: 13 });
+
+    return this.adapter.wait();
+  });
+
+  test("overrides static event options with dynamic event options", function(assert) {
```
magistrula: I struggled a bit with finding a brief way to differentiate the event options passed to `triggerable` and the options passed to the `triggerable` descriptor. Do `static` and `dynamic` get the right idea across?

magistrula: These would be another option:
```
calls Ember's triggerEvent helper with event options passed into `triggerable`
calls Ember's triggerEvent helper with event options passed into `triggerable` descriptor
overrides `triggerable` event options with `triggerable` descriptor event options
```

san650: I think the description is clear. I used to use the `at definition time` and `at execution time` to differentiate when defining the descriptor or executing the generated function. But I like the dynamic/static idea.

PR #317, opened 2017-06-11
Add getter macro

magistrula:
### Purpose
Provide a shorthand for creating descriptors like `baz` below:
```javascript
const page  = create({
  foo: isVisible('button'),
  bar: hasClass('active', 'button')

  baz: {
    isDescriptor: true,    
    get() {
      return this.foo && this.bar;
    }
  }
});
```

### Approach
Create a macro that takes a getter function and returns a descriptor:
  - The descriptor's `get` method returns the result of calling the passed-in getter function.
  - The getter function is executed with the descriptor's `this` context. It is not appropriate to pass in bound or arrow functions.
  - The getter is called with the property `key`, to maintain parity with the descriptor's `get` method.

### Test Coverage
Happy path & error handling covered by unit tests

### DIFF
--- a/addon/-private/properties/descriptor.js
+++ b/addon/-private/properties/descriptor.js
@@ -1,0 +1,58 @@
+import { throwBetterError } from '../better-errors';
+
+const NOT_A_FUNCTION_ERROR = 'Argument passed to `descriptor` must be a function.';
+
+/**
+ * Creates a descriptor whose value is determined by the passed-in function.
+ * The passed-in function must not be bound and must not be an arrow function,
+ * as this will prevent the function from being called with the correct context.
+ *
+ * @example
+ *
+ * // <input type="text">
+ * // <button>Submit</button>
+ *
+ * import { create } from 'ember-cli-page-object';
+ * import { descriptor } from 'ember-cli-page-object/macros';
+ *
+ * const page = create({
+ *   inputValue: value('input'),
+ *   isSubmitButtonDisabled: property('disabled', 'button'),
+ *
+ *   // with the `descriptor` macro
+ *   isFormEmpty: descriptor(function() {
+ *     return !this.inputValue && this.isSubmitButtonDisabled;
+ *   }),
+ *
+ *   // without the `descriptor` macro
+ *   _isFormEmpty: {
+ *     isDescriptor: true,
+ *     get() {
+ *       return !this.inputValue && this.isSubmitButtonDisabled;
+ *     }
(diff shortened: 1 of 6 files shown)


addon/-private/properties/descriptor.js
```diff
@@ -0,0 +1,55 @@
+import { throwBetterError } from '../better-errors';
+
+const NOT_A_FUNCTION_ERROR = 'Argument passed to `descriptor` must be a function.';
+
+/**
+ * Creates a descriptor whose value is determined by the passed-in function.
+ * The passed-in function must not be bound and must not be an arrow function,
+ * as this will prevent the function from being called with the correct context.
```
magistrula: In the `test` documentation (#318), I described the conditions in which an arrow function can be used successfully, but I can't think of a reason why an arrow function would ever be appropriate with `descriptor`...I can't imagine writing a `descriptor` that doesn't access page object properties via `this`. Thoughts?

addon/-private/properties/getter.js
```diff
@@ -0,0 +1,62 @@
...
+ * assert.ok(page.isFormEmpty);
+ *
+ * @public
+ *
+ * @param {Function} fn - determines what value is returned when the Descriptor is accessed
+ * @return {Descriptor}
+ *
+ * @throws Will throw an error if a function is not passed in as the first argument
+ */
+export function getter(fn) {
```
magistrula: @san650, I renamed `descriptor` as `getter` so that its usage is more evident from its name and so that the name reflects its purpose rather than its implementation.

addon/-private/properties/getter.js
```diff
@@ -0,0 +1,58 @@
...
+  return {
+    isDescriptor: true,
+
+    get(key) {
+      if (typeof fn !== 'function') {
+        throwBetterError(this, key, NOT_A_FUNCTION_ERROR);
+      }
+
+      return fn.call(this, key);
```
magistrula: I have just one question left, @san650. I am not wrapping `fn.call(this, key)` inside `executionContext.run` as in other helpers like `value` because I don't need access to the `context`. Do you think this is the correct approach, or would it be better to wrap this inside `executionContext.run` for the sake of consistency or in case some important additional logic is added to `run` in the future that is needed across all helpers?

san650: Wrapping it in `executionContext.run` is not needed. We can add it later if we change `executionContext.run` function.

PR #318, opened 2017-06-11
Add test macro

magistrula:
### Purpose
Provide a way to create a PageObject property whose value depends on whether some other property satisfies some set of criteria. For example, given the following PageObject:
```javascript
const page  = create({
  inputAutocomplete: property('autocomplete', 'input')
});
```
... we might want to define a property called `isAutocompleteEnabled`, which should be true if `inputAutocomplete` equals `on`.
```javascript
const page = create({
  inputAutocomplete: property('autocomplete', 'input'),
  isAutocompleteEnabled: test('inputAutocomplete', val => val === 'on')
});
``` 

### Test Coverage
Happy path & error handling covered by unit tests

### DIFF
--- a/tests/unit/-private/properties/alias-test.js
+++ b/tests/unit/-private/properties/alias-test.js
@@ -3,22 +3,26 @@ import {
   create,
   clickable,
   collection,
-  isVisible
+  isVisible,
+  text
 } from 'ember-cli-page-object';
-import { alias } from 'ember-cli-page-object/macros';
+import {
+  alias,
+  test as testMacro
+} from 'ember-cli-page-object/macros';
 
 moduleForProperty('alias', function(test) {
   test('can alias a top-level property', function(assert) {
     assert.expect(1);
 
     const page = create({
-      foo: isVisible('button'),
-      fooAlias: alias('foo')
+      isButtonVisible: isVisible('button'),
+      aliasedIsButtonVisible: alias('isButtonVisible')
     });
 
     this.adapter.createTemplate(this, page, '<button>Look at me</button>');
 
-    assert.ok(page.fooAlias);
+    assert.ok(page.aliasedIsButtonVisible);
   });
 
   test('can alias a top-level method', function(assert) {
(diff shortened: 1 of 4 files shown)


tests/unit/-private/properties/alias-test.js
```diff
@@ -3,31 +3,35 @@ import {
...
+  test as testMacro
+} from 'ember-cli-page-object/macros';
 
 moduleForProperty('alias', function(test) {
   test('can alias a top-level property', function(assert) {
     assert.expect(1);
 
     const page = create({
-      foo: isVisible('button'),
-      fooAlias: alias('foo')
```
magistrula: I made some naming tweaks to these tests for the sake of clarity/readability.

juanazam:
@magistrula I thinks this looks great! @san650 have discussed about something similar from the inception of this addon. We were not certain about having assertions inside a PO, but this is something in the middle. Having a testing function that's flexible enough to define which needs to be tested for a property and doing the assertions outside seems to be a great compromise.

tests/unit/-private/properties/alias-test.js
```diff
@@ -3,31 +3,35 @@ import {
...
   clickable,
   collection,
-  isVisible
+  isVisible,
+  text
 } from 'ember-cli-page-object';
-import { alias } from 'ember-cli-page-object/macros';
+import {
+  alias,
+  test as testMacro
```
magistrula: Having to rename `test` was a bit rough; I forgot to type `testMacro` several times when writing the tests below, and it took me a few minutes to figure out what I was doing wrong each time. Do we think this will cause any issues for people? Seems unlikely (because I'm not sure how you'd end up with two `test` values in a PageObject), but I wanted to raise the question.

magistrula:
@san650 & @juanazam, I'm closing this PR. After discussion with some other folks, I feel like there is unhelpful overlap between `test` and `getter` (#317). `getter` can accomplish the same thing `test` accomplishes...it's just slightly more verbose.

```javascript
// `test` macro
isButtonDisabled: test('buttonText', value => value === 'I am disabled')
```

```javascript
// `getter` macro
isButtonDisabled: getter(function() {
  return this.buttonText === 'I am disabled'
})
```

This overlap might be ok if the usage of `test` were really clear and obvious in a way that justified its existence, but I don't feel like that is the case. It doesn't feel like `test` needs to exist as a shorter version of `getter` above, but I do think that `equal` and `notEqual` should exist as a shorter version of using `getter` to check equality, so I'll follow up with that next.

Issue #220, opened 2016-07-27
predicate or query to check if an element is present on the page or not


skaterdav85:
Is there a predicate or query that checks if an element is present on the page or not? If not, is it something we can add? I'd be happy to send a PR if you agree.

san650:
@skaterdav85 we're using `isHidden` predicate to check if something is not visible on the page, this means if the element exists in the DOM but is hidden or if the element doesn't exist in the DOM.

From the user perspective the end result is the same, the user can't see the element, that's why we never had a helper to check if an element exists in the DOM.

That being said, I'm not against adding a helper to check this or to add an extra option to an existing helper. I would love to have the input of others as well. cc/ @juanazam @jeradg 

--

Creating a helper that checks for this is pretty straight forward

```js
import { findElement } from 'ember-cli-page-object';

export default function exists(selector, options) {
  return {
    isDescriptor: true,
    get() {
      return !!findElement(this, selector, options)
    }
}
```

With this implementation you get a lot of behavior for free: 

* Scope resolution works the same as for the rest of the helpers
* Adds a lot of the same options as for the rest of the helpers
  * @param {boolean} options.resetScope - Do not use inherited scope
  * @param {number} options.at - Filter by index using :eq(x) pseudo-class
  * @param {String} options.testContainer - Context where to search elements in the DOM

juanazam:
I'm 👍 on having this predicate, I have found myself with this problem a couple of times and thought might be handy to have it.

skaterdav85:
@san650 Thanks for the reply! Cool, didn't know that about isHidden. I'm fine with using isHidden then. I'll try and find this in the docs and update that.

YoranBrondsema:
I found myself needing this to detect the presence of a meta-tag. For a meta-tag, `is(':visible')` always returns `false`. I added an `exists` property in PR #323.

PR #323, opened 2017-08-01
Add `exists` accessor

YoranBrondsema:
Addresses #220.


san650:
@YoranBrondsema great addition to the library! I also found myself needing something like this. 

Thanks for contributing.

YoranBrondsema:
@san650 No problem, thanks for merging!

PR #324, opened 2017-08-02
Nested Page Objects

nikityy:


andrewfan:
@san650 hey, will this PR be merged?
having such syntaxis 
```
import ChildPO from 'ui/pods/components/current-plan/page-object';
import { create, nested } from 'ember-cli-page-object';

import {
  visitable
} from 'ember-cli-page-object';

const PO = {
  visit: visitable('global-settings/subscription'),
  child: nested(ChildPO, '.childSelector'),
};

export default PO;
```
vs 
```
import ChildPO from 'ui/pods/components/current-plan/page-object';
import { create as POcreate } from 'ember-cli-page-object';

import {
  visitable
} from 'ember-cli-page-object';

const PO = {
  visit: visitable('global-settings/subscription'),
};

export default PO;

export function create() {
  let po = POcreate(PO);
  po.child = ChildPO(CurrentPlanPO, '.childSelector');
  return po;
}
```
is a great benefit.

ro0gr:
@andrewfan I don't understand what is the benefit of `nested` approach over

```js
import { definition as childPODefinition } from 'ui/pods/components/current-plan/page-object';

import {
  visitable,
  create
} from 'ember-cli-page-object';

export const definition = {
  visit: visitable('global-settings/subscription'),
  child: childPODefinition,
};

export default create(definition);
```

?

ro0gr:
@nikityy canyon please motivate your change. What issue does it solve?

san650:
@nikityy @andrewfan we feel that using object composition and spread operator is enough to compose page objects. We want to avoid adding more complexity to the addon.

Maybe what we need is a section in the documentation explaining how to achieve this.

We can chat more about this on ember slack, channel #ec-page-object

Thanks.

Issue #311, opened 2017-05-19
Use native-dom-helpers internally


knownasilya:
https://github.com/cibernox/ember-native-dom-helpers

Relates to #256 and #289 

Leaving this as a placeholder.
cc @san650 @cibernox

san650:
Thanks @knownasilya for opening this issue.

ming-codes:
Well, just say the words on how you like to proceed. I'm sure there are people that like to contribute.

simonihmig:
Would also love to see this. Currently using `e-c-p-o` is the only reason we cannot drop jQuery from a new app we started with... 😥

One thing to keep in mind: all `e-n-d-h` helpers that trigger an event will return a promise (basically returning `wait()`), now usually used with `await`. This allows to get rid of some `wait()` / `andThen()` calls in the test itself, which is nice. 

But for this to work with `e-c-p-o`, all "actions" like `page.foo.click()` would also have to return this promise then, and not like now `this` for the fluent interface. So this would need to be a breaking change!

knownasilya:
You could still keep chainability, you'd just need to change the API to have an explicit method on the end to return the promise.

san650:
@simonihmig 
> But for this to work with e-c-p-o, all "actions" like page.foo.click() would also have to return this promise then, and not like now this for the fluent interface.

Since the beginning we made that all `e-c-p-o` nodes to _behave like a promise_ so you can use `await` (or `.then`) with actions.

All the following are supported right now

```js
const page = create({
  visit: visitable('...'),
  foo: clickable('...'),
  bar: clickable('...'),
});

await page.visit();
await page.visit().foo();
await page.visit().foo().bar();
await page.visit().foo().then(() => ...);
...
```

This is because of this

https://github.com/san650/ember-cli-page-object/blob/88fad42f23837020f40ffe57bfa2c56169aba39f/addon/-private/dsl.js#L12-L17

and this

https://github.com/san650/ember-cli-page-object/blob/88fad42f23837020f40ffe57bfa2c56169aba39f/addon/-private/dsl.js#L29

which is injected in all page objects nodes.

san650:
> Would also love to see this. Currently using e-c-p-o is the only reason we cannot drop jQuery from a new app we started with... 😥

[...]

Issue #289, opened 2017-03-22
[Feature Request] Remove jQuery dependency


leizhao4:
Related to #256 but in a broader sense. Since [Ember is going to make jQuery optional](https://github.com/emberjs/rfcs/pull/73), why not we use all vanilla JavaScript instead of relying on jQuery?

san650:
@leizhao4 this would be great! I think it could be an objective for the 2.0

I think right now the main dependency with jQuery are the custom jQuery selectors (`:contains`, etc.). We need to come up with a migration plan and start deprecating features that relay on jQuery.

juanazam:
@san650 rather than deprecating those features, it woud be great to have an optin way of using them if you install jQuery. Thoughts?

san650:
We could start reworking the internals to replace jQuery on places where it makes sense first. Like using native events for `triggerable` and `querySelector` and `querySelectorAll`.

ming-codes:
Would it be possible to configure jQuery as a “dev” dependency that’s only used in testing? That fulfills the requirement that we’re not shipping jQuery to production while still be able to use jQuery in tests.

ro0gr:
@ming-codes when you enable jQuery then Ember will immediately setup `Ember.$`. 
It's also possible to get [conflicts](https://github.com/san650/ember-cli-page-object/pull/325#issuecomment-333308426) with jQuery plugins. if they occasionally remained in your app.
Then when you disable jQuery for prod mode it would be impropely(I believe so) tested app.

If you are interested in reliable jQuery-less solution I would suggest to test #325. Here I've tried to properly isolate jQuery without a side effects.

PR #325, opened 2017-08-20
Integrate ember-native-dom-helpers

ro0gr:
addresses #311
reviewers @san650 @simonihmig 

`ember-native-dom` context allows not to rely on global ember test helpers(`click`, `fillIn`,..).
We still use `jquery` due to custom selectors capabilities(`:is`, `:contains` etc)

In order to allow ember application not to use `jquery` while keeping a backward compatibility `ember-cli-page-object` now bundles its [own jquery](https://github.com/san650/ember-cli-page-object/compare/master...ro0gr:native-dom-helpers?expand=1#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R91) copy namespaced as `'-jquery'`.

To enable native dom helpers in your tests, put this lines into your `tests/test-helper.js`:
```js
// tests/test-helper.js
...
import { useNativeDOMHelpers } from 'ember-cli-page-object/extend';

useNativeDOMHelpers(); // this will replace classic contexts with `e-n-d-h`
```

### Tests
I've expanded property tests with [native-dom] mode for both integration and acceptance contexts. It seems a little not optimal however it gives a solid code coverage with a minimal efforts.

### Todo
- [X] make properties tests pass for native contexts #327 
- [x] check[ if jquery is duplicated](https://github.com/san650/ember-cli-page-object/pull/325#discussion_r134330910).
- [x] ensure it doens't break jquery plugins.
 we use a [vendor shim](https://github.com/san650/ember-cli-page-object/pull/325/files#diff-f609b48b6e1d1ea0b3540d3787bd5266) to isolate jQuery from the app's one.
- confirm it works for real apps
  - [ ] with jquery in app
  - [ ] w/o jquery in app
  - [ ] fix jquery isolation - https://github.com/san650/ember-cli-page-object/pull/325#issuecomment-340882631
- [x] properly handle `endh` `keyEvent` https://github.com/cibernox/ember-native-dom-helpers/issues/112
- [ ] ~use `betterError` in integration context `visit`. [by @magistrula](https://github.com/san650/ember-cli-page-object/pull/325#discussion_r142842903)~
- [ ] make CI green (https://github.com/san650/ember-cli-page-object/pull/333)
- [x] update docs (https://github.com/san650/ember-cli-page-object/pull/334)
- [ ] rebase


addon/-private/execution_context/acceptance.js
```diff
@@ -75,25 +80,32 @@ AcceptanceExecutionContext.prototype = {
...
     return result;
   },
 
+  $(selector, container) {
+    if (container) {
+      return $(selector, container);
+    } else {
+      let testsContainer = this.testContext ?
+        this.testContext._element :
+        '#ember-testing';
```
ro0gr: This is definitely not the best thing to be hard-coded

addon/-private/execution_context/acceptance.js
```diff
@@ -75,25 +80,32 @@ AcceptanceExecutionContext.prototype = {
...
-    /* global find */
-    result = find(selector, options.testContainer || findClosestValue(this.pageObjectNode, 'testContainer'));
-
+    let result = this.$(selector, options.testContainer || findClosestValue(this.pageObjectNode, 'testContainer'));
     guardMultiple(result, selector, options.multiple);
 
     return result;
   },
 
+  $(selector, container) {
```
ro0gr: copied from integration context

ro0gr:
Actually `fillIn` helper is not used currently [due to support](https://github.com/cibernox/ember-native-dom-helpers/issues/100) of [`[contenteditable]`](https://github.com/san650/ember-cli-page-object/blob/88fad42f23837020f40ffe57bfa2c56169aba39f/addon/-private/execution_context/helpers.js#L21)

ro0gr:
Seems like `e-n-d-h` can absorb most of differences between acceptance and integration contexts.

index.js
```diff
@@ -11,6 +11,9 @@ module.exports = {
           enabled: this._shouldIncludeFiles(),
           import: ['index.js']
         };
+      },
+      jquery: {
+        import: ['dist/jquery.js'],
```
simonihmig: What happens here when your app has not opted out of using jQuery? Do we then bundle jQuery twice?

addon/-private/execution_context/acceptance.js
```diff
@@ -50,19 +56,18 @@ AcceptanceExecutionContext.prototype = {
       pageObjectKey: options.pageObjectKey
     });
 
-    /* global triggerEvent */
-    triggerEvent(selector, container, 'input');
-    triggerEvent(selector, container, 'change');
+    triggerEvent($selection, 'input');
+    triggerEvent($selection, 'change');
```
simonihmig: Ah, just saw your other comment, so ignore this...

simonihmig:
IIUC you have jQuery still around as a compatibility layer for jQuery-specific selectors like `:eq()`, right? I cannot remember that I have ever used this in tests. So what about the following other approach:

* we keep the existing execution contexts around basically unchanged
* we add additional ones (*maybe even unified for acceptance and integration?*) based on e-n-d-h, that do *not* require jQuery at all. Obviously non-standard selectors will not work with them. Users have a way to opt in into the new native execution contexts, e.g. by setting some option in `ember-cli-build.js`, so some broccoli funnel replaces the default context files with those new native-based ones. 

This way 
* we could keep the existing ones untouched, reducing the risk to break something
* the implementation of the new native contexts could focus on a native-only implementation, making this easier to reason about. And they would not need to be backwards compatible!
* these new ones could land with some experimental support, for users to try this out 
* once we are happy with them, we can declare them stable, and add some deprecation message to the old ones
* with a major version bump the old ones could be dropped and the new ones become the new standard

What's your thoughts on this? @ro0gr @san650 

cc @cibernox interested in chiming in here?

index.js
ro0gr: This needs to be verified. 
Anyways seems like it's not a big deal cause it's for testing only.
Also this way we can guarantee that we don't consume jquery from ember

ro0gr:
@simonihmig thank you for your thoughts.

> IIUC you have jQuery still around as a compatibility layer for jQuery-specific selectors like :eq(), right?

Yes. I think all the specific selectors can be found in [Selector class](https://github.com/san650/ember-cli-page-object/blob/88fad42f23837020f40ffe57bfa2c56169aba39f/addon/-private/helpers.js#L38)

> we add additional ones (maybe even unified for acceptance and integration?) based on e-n-d-h, that do not require jQuery at all

Unfortunately I think it's impossible right now. Lots of [`properties`](https://github.com/san650/ember-cli-page-object/blob/88fad42f23837020f40ffe57bfa2c56169aba39f/addon/-private/properties/fillable.js#L128) rely on `buildSelector()` which builds [selectors](https://github.com/san650/ember-cli-page-object/blob/88fad42f23837020f40ffe57bfa2c56169aba39f/addon/-private/helpers.js#L38) with jq like filters. It means that with the native context getting of nth item from [`collection`](https://github.com/san650/ember-cli-page-object/blob/88fad42f23837020f40ffe57bfa2c56169aba39f/addon/-private/properties/collection.js#L41) will not work.

> we could keep the existing ones untouched

nah 😋   Currently I'm working on merging contexts. Seems like `e-n-d-h` absorbs most of the differences between integration/acceptance modes.

ro0gr:
Last commit contains unification of contexts. both of them are derived from base [execution context](https://github.com/ro0gr/ember-cli-page-object/blob/a55c79298fa0e67411e5aeea3b933de5edcde963/addon/-private/execution_context/context.js). The only difference is a `runAsync` method:
 - [acceptance](https://github.com/ro0gr/ember-cli-page-object/blob/a55c79298fa0e67411e5aeea3b933de5edcde963/addon/-private/execution_context/acceptance.js#L9)
 - [integration](https://github.com/ro0gr/ember-cli-page-object/blob/a55c79298fa0e67411e5aeea3b933de5edcde963/addon/-private/execution_context/integration.js#L13)

I'm not sure if e-n-d-h can handle this now.

@san650 Now we try to run the same set of [property tests](https://github.com/ro0gr/ember-cli-page-object/blob/a55c79298fa0e67411e5aeea3b933de5edcde963/tests/helpers/properties.js) against all the contexts. As I mentioned in pr description there are some issues with these tests.
I think merging of contexts can allow us to simplify the test suite a little bit because we might not need to care about specific context anymore. Does it sound reasonable to you?

ro0gr:
I agree with @simonihmig that leaving classical contexts untouched is a good thing. So I've reverted all the changes in classic contexts with global ember helpers. 

In order to enable `e-n-d-h` now you can use `useNativeDOMHelpers`:
```js
// tests/test-helper.js
...
import { useNativeDOMHelpers } from 'ember-cli-page-object/extend';

useNativeDOMHelpers(); // this will replace classic contexts with `e-n-d-h`
```
Currently it doesn't accept any args.

> Users have a way to opt in into the new native execution contexts, e.g. by setting some option in ember-cli-build.js, so some broccoli funnel replaces the default context files with those new native-based ones.

I've thought about build time config for this. But I think this way it will be harder to test all the contexts inside a single `e-c-p-o` test suite. So I've chosen run-time switch in `test-helper.js`

index.js
ro0gr: I've just used this branch on a project with `jquery@1.11.3` included. As a result both `jquery` versions are bundled into `vendor.js`.

The interesting thing is that `1.11.3` is still used by `ember` and `v3.2.1` by `e-c-p-o` so their are no in conflict.

ro0gr: all the jquery plugins should be included before `e-c-p-o`. Otherwise they will extend `e-c-p-o`'s jquery instead of `Ember.$` :/

ro0gr:
In the last commit I've tried to import `ember-cli-page-object`s jquery instance in the very beginning of `vendor.js`. 
I believe it would allow jquery plugins to extend `Ember.$` because `Ember` now comes later in the `vendor.js`.

It would be awesome if someone who has an ember app with several jquery plugins and jquery itself can check their test suite against this branch 👼:
```sh
npm i --save-dev https://github.com/ro0gr/ember-cli-page-object#native-dom-helpers
```

ro0gr:
Let's assume the following:
 - `jquery` is excluded from the app
 - you have installed this branch of `ecpo` (before "[truly isolate jquery](https://github.com/san650/ember-cli-page-object/pull/325/commits/2ae7e7b8f433d2d32553f9a5959b22a16eea8712)" commit) 
 - some sly `jquery` plugin is left in your code base **by accident**

As a result `ember s` and `ember t` will work because this `jquery` plugin is able to use `jquery` from `ecpo`.
At the same time staging/production builds are getting unusable as long as `ecpo` is stripped from production builds. 

I've ended up in similar situation. [truly isolate jquery](https://github.com/san650/ember-cli-page-object/pull/325/commits/2ae7e7b8f433d2d32553f9a5959b22a16eea8712) tries to address this issue by calling `.noConflict()` and namespacing `jquery` as a pivate module `'-jquery'`. This trick is happening [right after local `jquery` version is imported](https://github.com/san650/ember-cli-page-object/pull/325/commits/2ae7e7b8f433d2d32553f9a5959b22a16eea8712#diff-168726dbe96b3ce427e7fedce31bb0bcR34).

tests/helpers/properties.js
```diff
@@ -1,37 +1,56 @@
...
 export function moduleForProperty(name, cbOrOptions, cb) {
   let options = cb ? cbOrOptions : {};
   cb = cb || cbOrOptions;
 
-  // Generate acceptance tests
-  moduleForAcceptance(`Acceptance mode | Property | ${name}`, {
-    beforeEach() {
-      this.adapter = new AcceptanceAdapter(AcceptanceExecutionContext);
-    },
+  [true, false].forEach(_useNativeDOMHelpers => {
```
ro0gr: I'm not a big fan of this. But it gives us an ability to test `native-dom-helpers` almost for free. So we can use existing test suite a baseline.
I have few ideas on how to improve testing story, but I prefer to leave it for a separate issue/pr.

index.js
```diff
@@ -23,6 +30,11 @@ module.exports = {
 
     this.app = app;
 
+    if (this._shouldIncludeFiles()) {
+      this.import('vendor/shims/-jquery.js', { prepend: true });
+      this.import('vendor/ecpo-jquery/dist/jquery.js', { prepend: true });
```
ro0gr: ensure shims goes right after our private `jquery` instance

vendor/shims/-jquery.js
```diff
@@ -0,0 +1,11 @@
+(function() {
+  var jq = self['$'].noConflict();
```
ro0gr: here we heavily rely on the assumption that we included jquery right a moment ago 
https://github.com/san650/ember-cli-page-object/pull/325/files#diff-168726dbe96b3ce427e7fedce31bb0bcR35

this way `noConflict` gives us our jquery instance and disposes it from the `window`.

ro0gr: I think I'll inline some comments in code a little bit later

addon/-private/execution_context/integration.js
```diff
@@ -33,7 +33,9 @@ IntegrationExecutionContext.prototype = {
   },
 
   // Do nothing in integration test
-  visit: $.noop,
+  visit() {
+    throw new Error('visit is not supported in integration mode');
```
ro0gr: should I rollback this?

ro0gr:
Currently all the `trigger` tests which test keypress `13` are failing on CI in `native-dom-helpers` mode.
I see similar issue with IE currently. At the same time it works perfectly with chrome and firefox.

I'm sure the issue originated from [`fireEvent`](https://github.com/cibernox/ember-native-dom-helpers/blob/9132376a0f4e9e2529716016f36b65be1321ec5d/addon-test-support/fire-event.js#L120). 
The next step is to fix this.

addon/-private/execution_context/context.js
```diff
@@ -0,0 +1,131 @@
+import $ from '-jquery';
```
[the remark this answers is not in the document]
ro0gr: yes I also thought about it. 
Actually I like `dom-context` more. `native` sounds misleading to me a little bit. Maybe `native-dom-context` is more appropriate here?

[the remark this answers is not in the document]
magistrula: `native-dom-context` sounds good to me!

addon/extend.js
```diff
@@ -1,4 +1,21 @@
 export { findElement } from './-private/extend/find-element';
 export { findElementWithAssert } from './-private/extend/find-element-with-assert';
 export { buildSelector, getContext, fullScope } from './-private/helpers';
-export { register as registerExecutionContext } from './-private/execution_context';
+import { register as registerExecutionContext } from './-private/execution_context';
+
+import IntegrationNativeContext from './-private/execution_context/integration-native';
+import AcceptanceNativeContext from './-private/execution_context/acceptance-native';
+import IntegrationClassicContext from './-private/execution_context/integration';
+import AcceptanceClassicContext from './-private/execution_context/acceptance';
```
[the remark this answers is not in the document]
ro0gr: > Suggest `IntegrationJQueryContext` and `AcceptanceJQueryContext` to increase clarity. 

Currently `AcceptanceNativeContext ` and `IntegrationNativeContext`  both use `jquery` to find elements.

I'd suggest to rename all the context's module names. 
How about:
- `acceptance-ember-context`, 
- `integration-ember-context` 
- `acceptance-native-dom-context`
- `integration-native-dom-context` 

It seem too verbose to me. But I'm ok with it.

addon/-private/execution_context/context.js
```diff
@@ -0,0 +1,131 @@
...
+    triggerEvent(el, 'input');
+    triggerEvent(el, 'change');
+  },
+
+  $(selector, container) {
+    if (container) {
+      return $(selector, container);
+    } else {
+      let testsContainer = this.testContext ?
+        this.testContext._element :
```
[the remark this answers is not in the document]
ro0gr: '#ember-testing' [one line above](https://github.com/san650/ember-cli-page-object/pull/325/files/01d28c6645250fad0eb08a8a71b68103f259f99d#diff-48f213875b681062c2ceb1e789634adbR63) also doesn't look very nice.

addon/-private/execution_context/context.js
```diff
@@ -0,0 +1,131 @@
...
+  // Do nothing in integration test
+  visit: visit,
+
+  click(selector, container) {
+    const el = this.$(selector, container)[0];
+    click(el);
+  },
+
+  fillIn(selector, container, options, content) {
+    let el = this.$(selector, container)[0];
```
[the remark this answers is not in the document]
ro0gr: I've never used `fillIn` with `{ multiple: true }` :). I'll update it. thanks.

addon/extend.js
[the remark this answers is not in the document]
magistrula: How about `acceptance-jquery-event-context`, `acceptance-native-event-context`, etc? And then, `native-event-context` instead for `context.js` above? If the file names are too long, we could add `acceptance` and `integration` folders, each of which could include `native-event-context` and `jquery-event-context`.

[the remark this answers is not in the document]
ro0gr: > How about acceptance-jquery-event-context
- i'm not sure why do you refer to `event`, triggering of events is only a part of context
- `jquery` is not dedicated only for existing contexts. All of contexts(including `native-dom`)  use jquery to find elements to find DOM elements. More details [here](https://github.com/san650/ember-cli-page-object/issues/289#issuecomment-289348460)

> native-event-context instead for context.js above?

yes, sure. Anyway I think we agreed on [`native-dom-context`](https://github.com/san650/ember-cli-page-object/pull/325#discussion_r143362645) name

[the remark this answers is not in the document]
magistrula: It seems that the core difference between these contexts is the way in which the events are triggered, so that's why I'm suggesting filenames with `event` in them as well as `native-event-context` instead of `native-dom-context`. Not a big deal either way...just pondering how to achieve the most clarity.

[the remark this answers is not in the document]
ro0gr: Now I see your point about events. However `acceptance-jquery-event-context` is still not explanatory enough cause it also uses global ember test helpers(`find`, `click`). I'm also agree with you that it's not a big deal cause it's not a part of public API.
I think it might be better not too touch exiting context names at all until a better name eventually comes to us :)

addon/-private/execution_context/context.js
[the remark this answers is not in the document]
ro0gr: According to [`fillable` API docs](http://ember-cli-page-object.js.org/docs/v1.12.x/api/fillable) there is no support for `multiple` option. Does `fillable` intend to support this option?

ro0gr:
hm.., I've just checked IE and all the tests started to pass after `keyEvent` fix event in IE9 emulation mode. However CI with `phantomjs` still fails. 

I'm not sure if we want to fix issues in `phantomjs` because it's abandoned in community now...

addon/-private/execution_context/integration.js
ro0gr: I've tried to use `throwBetterError` here. Unfortunately it doesn't work out of the box because `throwBetterError` is currently coupled elusively to element find failures.

 This pr is already quite large. I've reverted original silent behavior.

ro0gr:
I think the best way to avoid issue with `phantomJS` is to [replace it](https://github.com/san650/ember-cli-page-object/pull/333) with another browser

addon/-private/execution_context/context.js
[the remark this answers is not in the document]
magistrula: You're right that it's not called out in the docs, but `fillable` does work with `{ multiple: true }`, so it's probably safest to preserve that functionality at this point.

addon/-private/execution_context/integration.js
magistrula: For future reference, here's an example of where `throwBetterError` is used outside the context of element find failures: https://github.com/san650/ember-cli-page-object/blob/master/addon/-private/properties/alias.js#L86.

addon/-private/execution_context/context.js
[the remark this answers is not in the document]
ro0gr: OK, will do. Thx

addon/-private/execution_context/native-dom-context.js
```diff
@@ -0,0 +1,138 @@
...
+  click(selector, container) {
+    const el = this.$(selector, container)[0];
+    click(el);
+  },
+
+  fillIn(selector, container, options, content) {
+    let elements = this.$(selector, container).toArray();
+
+    elements.forEach((el) => {
+      fillElement(el, content, {
```
magistrula: Todo: 
1. Update function signature of `fillElement` to indicate that the first param is an element
2. Update the usage of `fillElement` in the other places where it is used so that we are always passing an element and not a jQuery selection.

index.js
magistrula: Should we put a comment here indicating that this ordering is important for `shims/-jquery` to work? The comment in `shims/-jquery` is really helpful, and I think the assumption we're making there about ordering will be more secure if there is also a warning here, where that ordering is defined.

pzuraq:
I tried this branch locally to see if it would work (we use native dom events so this is going to be necessary for us) and I'm getting `define is not a function`, appears that the jquery shim file is getting prepended before `define` is actually assigned to be function. This is with `ember-cli@2.14.0`

ro0gr:
@pzuraq thank you for report. I've tested it against ember-cli@2.12 and 2.16 and haven't noticed such issue yet. 

I'll test it with 2.12. However some more additional details would be helpful:
 - what's you npm version
 - is it happening inside an app or addon
 - do you do npm link or just specify a git URL in your package.json

Thanks

pzuraq:
* We're using `yarn@v1.1.0` 
* This is happening inside an addon 
* I'm using a git URL in package.json

addon/-private/execution_context/acceptance-native-dom.js
```diff
@@ -0,0 +1,23 @@
...
+}
+
+AcceptanceExecutionContext.prototype = Object.create(ExecutionContext.prototype);
+
+AcceptanceExecutionContext.prototype.visit = function() {
+  return visit(...arguments);
+};
+
+AcceptanceExecutionContext.prototype.runAsync = function(cb) {
+  window.wait().then(() => {
```
pzuraq: This appears to be the last dependency on a global test helper, which means we still have to use `setContext` to be able to use PageObjects in integration tests. We can do the same thing `ember-native-dom-helpers` does here to make this completely independent:

```js
import wait from 'ember-test-helpers/wait';

(window.wait || wait)().then(() => {});
```

ro0gr: You are right , seems like it should be fixed.

ro0gr:
@pzuraq I've updated handling of jquery per you suggestion in slack. 
Now we store it in a temporary global variable in the very beginning and later expose it as an amd module.

ro0gr:
superseded by #339

PR #327, opened 2017-08-29
change property unit tests to assert against real DOM

ro0gr:
Current approach for testing properties doesn't scale very well. For example it's impossible to run existing properties tests against alternative contexts like [native contexts](https://github.com/san650/ember-cli-page-object/pull/325).

The general idea of this pr is changing assertions like
```js
this.adapter.click((actualSelector) => assert.equal(actualSelector, expectedSelector));
```
to
```js
  this.adapter.$(expectedSelector).on('click', () => assert.ok(true));
```
This way we don't test internals of contexts anymore.

@san650 There is some work left to do here(mostly cleanup). Anyway I would appreciate your thoughts on this change. Does this approach make sense to you?

### DIFF
--- a/tests/unit/-private/properties/click-on-text-test.js
+++ b/tests/unit/-private/properties/click-on-text-test.js
@@ -234,6 +234,22 @@ moduleForProperty('clickOnText', function(test) {
     }, /page\.foo\.bar\.baz\.qux/, 'Element not found');
   });
 
+  test("doesn't raise an error when the element is not visible and `visible` is not set", function (assert) {
+    assert.expect(1);
+
+    let page = create({
+      foo: clickOnText('button')
+    });
+
+    this.adapter.createTemplate(this, page, '<button style="display:none">Click me</button>');
+
+    this.adapter.$('button').on('click', () => assert.ok(1));
+
+    page.foo('Click me');
+
+    return this.adapter.wait();
+  });
+
   test('raises an error when the element is not visible and `visible` is true', function(assert) {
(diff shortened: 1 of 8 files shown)


tests/unit/-private/properties/click-on-text-test.js
```diff
@@ -258,24 +234,6 @@ moduleForProperty('clickOnText', function(test) {
     }, /page\.foo\.bar\.baz\.qux/, 'Element not found');
   });
 
-  test("doesn't raise an error when the element is not visible and `visible` is not set", function(assert) {
```
ro0gr: removed by mistake

tests/unit/-private/properties/fillable-test.js
```diff
@@ -13,62 +13,77 @@ moduleForProperty('fillable', function(test) {
...
+    '<select name="clue"><option></option><option>dummy text</option></select>',
+    '<select id="clue"><option></option><option>dummy text</option></select>',
+  ];
+
+  formControlTemplates.forEach(template => {
+    let gtPos = template.indexOf('=');
+    let name = template.substr(1, gtPos - 1);
+    const [tagName, attrName] = name.split(' ');
+
+    test(`looks for ${tagName} with ${attrName}`, function(assert) {
```
ro0gr: need to figure out how to simplify `[contenteditable]`, `input`, `textarea`, `select` tests

san650:
Thanks for simplify the test @ro0gr

PR #334, opened 2017-10-18
Add native-dom mode documentation

ro0gr:
I guess we're gonna to release `native-dom` mode in 1.13 so I've copy-pasted pages from **v1.12.x** to **v1.13.x**.

Also a simple page describing why do we need `ember-native-dom-helpers` and how to enable it was added.

I'm open for any suggestions on re-phrasing or any improvements of the new page cause I'm not very good in gluing the words 😄 

topics to cover:
- [x] wait for #339 to be merged to stabilize API 
- [x] explain ember-native-dom-event-dispatcher
- [x] jquery
  - ~~ember@2.13.0-beta.2~~
  - events
  - ~~[isVisible](https://github.com/san650/ember-cli-page-object/pull/334#issuecomment-345399424)~~ #346

### DIFF
--- a/docs/v1.13.x/native-dom.md
+++ b/docs/v1.13.x/native-dom.md
@@ -1,0 +1,28 @@
+---
+layout: page
+title: Native DOM Helpers
+---
+
+{% raw %}
+By default `ember-cli-page-object` uses global ember test helpers such as `click`, `fillIn`, `find`, etc.
+While it works great this approach has one downside: global ember test helpers require `jQuery` to be bundled within your `Ember` global.
+
+As a result if you want to drop a dependency on `jQuery` in your app or addon you won't be able to use standard `Ember` test helpers.
+
+In order to solve this problem `ember-cli-page-object` provides an integration with [`ember-native-dom-helpers`](https://github.com/cibernox/ember-native-dom-helpers).
+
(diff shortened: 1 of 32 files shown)


ro0gr:
tbh I haven't ruby installed locally 🙊 ... so I haven't tried it locally. 
I'm curious if there is a way to test resulting docs instantly?

docs/v1.13.x/native-dom.md
```diff
@@ -0,0 +1,28 @@
+---
+layout: page
+title: Native DOM Helpers
```
magistrula: Suggest `Triggering Native vs. jQuery DOM Events` as the title of this page. Would be nice to have something as expressive as possible so that people can easily realize "oh, yeah...that sounds like what I'm dealing with."

ro0gr: Seems like it's a good direction. I'm finally [agree](https://github.com/san650/ember-cli-page-object/pull/339#issuecomment-344063047) with you that the key thing here are Events

ro0gr:
from slack by @magistrula 
> in jquery event mode `isVisible` returns `false` for an empty div, but in native event mode, `isVisible` returns `true` for an empty div.

magistrula:
@ro0gr I'm going to close this PR, since these documentation updates were handled in this week's release. Thanks for supplying the initial draft of the documentation for native-events mode!

PR #336, opened 2017-10-22
Fixes CSS group (,) selector in scope

### LINKED ISSUE #335: Selector scope concatenation does not work with comma selector
```
create({
  scope: '.level-1', 

  nested: {
    scope: '.one, .two'
  }
})
```

Expected: `.level-1 .one, .level-1 .two`

Actual: `.level-1 .one, .two`

ming-codes:
Closes #335

### DIFF
--- a/addon/-private/helpers.js
+++ b/addon/-private/helpers.js
@@ -60,7 +60,7 @@ class Selector {
     scopes.reverse();
     scopes.push(targetScope);
 
-    return $.trim(scopes.join(' '));
+    return $.trim(fullSelector(scopes));
   }
 
   getScopes(node) {
@@ -71,11 +71,15 @@ class Selector {
     }
 
     if (!node.resetScope && Ceibo.parent(node)) {
-      scopes = scopes.concat(this.calculateScope(Ceibo.parent(node)));
+      scopes = scopes.concat(this.getScopes(Ceibo.parent(node)));
     }
 
     return scopes;
   }
+}
+
+function splitSelector(selector) {
+  return selector.split(/\s*,\s*/); // naive impl
 }
 
 export function guardMultiple(items, selector, supportMultiple) {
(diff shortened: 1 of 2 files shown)


juanazam:
@ming-codes This looks good! I will give it a try to see if everything works correctly, thanks for your contribution!

magistrula:
@ming-codes, I'm wondering...what's the use case for defining a scope that targets elements with two different selectors rather than giving both elements a common selector? This change introduces some nice flexibility, but just wanted to do a philosophical gut-check.

juanazam:
@magistrula a few months ago I has a case that applied for this feature:

Imagine you have a list of items, user avatars, when a user has an avatar, display image, if the user doesn't have one, display initials.

```
  <ul class="user-bubbles">
    <li class="user-avatar"><img src={{user1.avatarUrl}}/></li>
    <li class="user-initials">{{user.initials}}</li>
  </ul>
```
And now in a test you want to count the number of users present in that list:

```
const page = {
  users: { scope: '.user-vatar,.user-initials' }
}
assert.equal(page.users.count, 2)
```

What do you thing?

magistrula:
@ming-codes I see what you're getting at. I think something like this could also work for your use case:
```hbs
<ul class="user-bubbles">
  <li class="user-avatar" data-test-user-bubble><img src={{user1.avatarUrl}}/></li>
  <li class="user-initials" data-test-user-bubble>{{user.initials}}</li>
</ul>
```
```javascript
const page = {
  users: {
    scope: '[data-test-user-bubble]'
  }
}
```
If those list items are basically two versions of the same thing, then it would make sense for them to have some attribute that says "hey, we're both user info bubbles." Currently, since `scope` does not support `,` your only option right now is to go ahead and add some common attribute so that you can target both list items with one selector. If we add support for `,` in `scope`, then you wouldn't have to add a common attribute, and I'm wondering if that will lead to less clarity overall. So basically, I'm torn over whether this flexibility might be in conflict with other best practices.

ming-codes:
@magistrula My use case for `,` selector was that I have a component that has multiple child components that have similar grandchildren structure. So I wanted to use `collection` helper to be able to iterate through multiple children and map out the grandchildren structure too. Since the child level components are different, `,` make sense in the case.

I think your workaround to use another common class can work for me too.

juanazam:
@ming-codes Hey, after discussing this over slack with the rest of the `ember-cli-page-object` team we decided not merge this. We believe it has valid use cases but it also introduces easy ways making a page object hard to understand. There are other ways of working this out, for example using a test selector with `ember-test-selector`.

Again, thanks for taking time to look into this and sorry about the bad news.

Cheers!

PR #338, opened 2017-11-08
Small fix to documentation for `isPresent`

YoranBrondsema:
Also made a small improvement to `docs.js` to create the `gh-branch` if it doesn't exist yet.


san650:
Thanks Yoran!

PR #333, opened 2017-10-17
Update ember-cli@2.16

ro0gr:
I've used [ember-cli-update](https://github.com/kellyselden/ember-cli-update) for automatic update.
It also ran[ ember-modules-codemod](https://github.com/ember-cli/ember-modules-codemod) to update Ember imports.

Notable changes:

- new modules imports
- jshint -> eslint
- PhantomJS -> Chrome/Chrome Headless for CI
- No usages of Ember.HTMLBars.compile and page.render. It replaced by `htmlbars-inline-prcompile` and `this.render()`.

[Originally](https://github.com/san650/ember-cli-page-object/pull/332) I've tried to remove bower dependencies as well, but it led to test failures for ember@1 scenarios. I think it can be solved in a separate pr by proper ember try configuration.

Also we've added `testdouble` dependency for now in order to make node tests pass.

PR #339, opened 2017-11-11
Add native-dom mode

ro0gr:
addresses #311 

I've rebased commits from #325 for more convenient review and addressed @pzuraq [comment](https://github.com/san650/ember-cli-page-object/pull/325#discussion_r148116094).

---

`ember-native-events` context allows not to rely on global ember test helpers(`click`, `fillIn`,..).

One of the benefits of native-events mode is an ability to test jquery-free applications.
However we still use jquery selectors like `:is`, `:contains` etc(#289). 
In order to drop dependency on jquery provided by Ember by default we bundle our own jquery version.

To enable native dom helpers in your tests, put this lines into your `tests/test-helper.js`:
```js
// tests/test-helper.js
...
import { useNativeEvents } from 'ember-cli-page-object/extend';

useNativeEvents(); // this will replace classic contexts with `e-n-d-h`
```

### Tests
I've expanded property tests with [native-dom] mode for both integration and acceptance contexts. It seems a little not optimal however it gives a solid code coverage with a minimal efforts.

### todo
- [ ] make CI green (https://github.com/san650/ember-cli-page-object/pull/333)


ro0gr:
As @pzuraq noticed in #ec-page-object slack channel we still allow to specify jquery selectors in page objects in a native-dom mode:

> [7:26 PM] moreso what I mean is here: https://github.com/san650/ember-cli-page-object/pull/339/files#diff-66f0a7ca63b50a6b9973d119ab2570a5R43
> [7:26] in theory, a user could pass in a jquery selector
> [7:27] which would then be matched correctly, pull the corresponding element out, and apply the native dom helper
</details>

That means the current native-dom name is not very precise. We might get a naming collision If we occasionally will decide to provide true native-dom mode w/o jquery on board. So in the last commit I've  renamed native-dom to native-events as @magistrula [previously suggested](https://github.com/san650/ember-cli-page-object/pull/325#discussion_r144400773).

Also `useNativeDOMHelpers` renamed to `useNativeEvents`

san650:
@ro0gr can you rebase with master?

ro0gr:
Sure. Will do it today

ro0gr:
It's finally green 🎉

san650:
👏 👏 👏 👏 🎉 🍾

simonihmig:
@ro0gr Thanks a lot for working on this! Also to all others who helped to get this to land! (@pzuraq @magistrula @san650) This is highly appreciated! 👍 👏

I just wished I could have helped more, but time was too limited. But we do have an app I want to give the alpha version a try, will report back if there are any issues!

PR #340, opened 2017-11-14
bugfix(nested-usage): Fixes nested addon usage by finding the host

### LINKED ISSUE #337: Appears to break when included in nested addons
It appears that the inclusion strategy in `treeFor` doesn't work when included in nested addons. I haven't been able to figure out exactly why, but I can't seem to get the correct context from anywhere.

We could switch to using a similar strategy to [ember-native-dom-helpers](https://github.com/cibernox/ember-native-dom-helpers/blob/master/index.js#L7) to provide the same external API, but not determine whether or not to include things ourselves. Let `ember-cli` decide based on public hooks.

pzuraq:
Only addons that are immediately included at the root level app actually get access to the app, but addons can also traverse their ancestry to find that root level addon. This adds a function based on the private [`findHost` function in ember-cli itself](https://github.com/ember-cli/ember-cli/blob/master/lib/models/addon.js#L613). While the method is private, other addons such as ember-test-selectors, ember-bootstrap, and [plenty of others](https://emberobserver.com/code-search?codeQuery=findHost) do use it, and since it's based on the tree structure that addons have it's pretty unlikely that it will break on us soon.

I can confirm that this fixes #337 locally, happy to try to add a test in node-tests as well if needed

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,13 +15,8 @@ module.exports = {
     }
   },
 
-  included: function(app) {
-    // see: https://github.com/ember-cli/ember-cli/issues/3718
-    if (typeof app.import !== 'function' && app.app) {
-      app = app.app;
-    }
-
-    this.app = app;
+  included: function() {
+    this.app = this._findHost();
 
     this._super.included.apply(this, arguments);
   },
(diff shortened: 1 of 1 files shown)


ro0gr:
I'm interested what happens if several addons includes `ember-cli-page-object`. Is it duplicated in a target app? which one wins?

san650:
@pzuraq this fix is released in v1.12.1. Thanks for contributing!

Issue #202, opened 2016-06-20
`visitable` doesn't encode dynamic segments


san650:
e.g.

``` js
const page = create({
  visit: visitable('/users/:id')
});

page.visit({ user: 'a/user' }); // => visits('/users/a/user');
```

The expected behavior is to navigate to `/users/a%2Fuser'` instead of `/users/a/user`

san650:
I think we have to `encodeURIComponent` every dynamic segment value

san650:
Fixed by #342

PR #342, opened 2017-11-16
#202 visitable doesn't encode dynamic segments bug fix

antoarundominic:


san650:
Thanks for contributing!

Issue #288, opened 2017-03-06
Add ability to look up items in a collection via properties other than index


lsthornt:
Use case:
```js
<ul>
  <li data-test-id='UUID1'>Object 1</li>
  <li data-test-id='UUID2'>Object 2</li>
</ul>

PageObject.create({
  objects: collection({
    scope: 'ul',
    itemScope: 'li'
  })
});

page.objects("[data-test-id='UUID1']");
```
---

This can currently be solved by using nested PageObjects, a la:
```js
PageObject.create({
  objects(uuid) {
    return PageObject.create({
      scope: `[data-test-id='${uuid}']`
    });
  }
});

page.objects('UUID1');
```
However there are some drawbacks to this, including the fact that the outer PageObject's scope will not be inherited.

ro0gr:
https://github.com/san650/ember-cli-page-object/issues/198 related.

dadleyy:
I didn't want to make a completely separate issue for this, but I was wondering if it would be niceto have collections be able to receive a list of indices to return:

```js
const [a,b,c] = page.objects([1,2,3])
vs.
const [a,b,c] = page.objects().toArray();
```

dodeja:
@lsthornt Interesting. How does this work with component integration tests?

pzuraq:
I think this could actually more safely be accomplished by allowing collections to query the page objects of the items in the collections themselves, rather than directly querying the items:

```js
<ul>
  <li data-test-id='UUID1'>Object 1</li>
  <li data-test-id='UUID2'>Object 2</li>
</ul>

PageObject.create({
  objects: collection({
    scope: 'ul',
    itemScope: 'li',

    item: {
      id: attribute('data-test-id')
    }
  })
});

page.objects('id', 'UUID1');
```

This way you aren't breaking encapsulation, the item in the collection gets to define it's own content and values you can search. It could be a completely separate component definition, and you would essentially be searching its public API

ro0gr:
I think we should have a way to describe where an element id can be found. So `id` field sounds like a good candidate for a default collection DSL to me. If we have an `id` we then also need something like `getById`.

```js
const form = create({
  fields: collection({
    scope: 'input',
    id: attribute('name') 
  })
});

const login = form.fields.byId('login')
const password = form.fields.byId('password')
``` 

[...]

PR #345, opened 2017-11-18
feature(collection): Adds find and findBy array methods

pzuraq:
This PR adds the `find` and `findBy` array methods, which make searching for a single item in an array much more convenient. It's a common use case that we want to search a list of items for some item that matches a particular criteria (`fullName === 'John Doe'`, etc.) and while we can currently do it with `filterBy`, it's just a little extra work to pull it out.

### DIFF
--- a/addon/-private/properties/collection.js
+++ b/addon/-private/properties/collection.js
@@ -5,7 +5,7 @@ import { create } from '../create';
 import { count } from './count';
 import Ceibo from 'ceibo';
 
-const arrayDelegateMethods = ['map', 'filter', 'mapBy', 'filterBy', 'forEach'];
+const arrayDelegateMethods = ['map', 'filter', 'find', 'mapBy', 'filterBy', 'findBy', 'forEach'];
 
 function merge(target, ...objects) {
   objects.forEach((o) => mergeFunction(target, o));
--- a/tests/unit/-private/properties/collection-test.js
+++ b/tests/unit/-private/properties/collection-test.js
@@ -99,6 +99,10 @@ moduleForProperty('collection', function(test) {
 
     assert.deepEqual(page.foo().filter((i) => i.isSpecial).map((i) => i.text), ['Lorem']);
     assert.deepEqual(page.foo().filterBy('isSpecial').map((i) => i.text), ['Lorem']);
+
+    assert.deepEqual(page.foo().find((i) => i.isSpecial).text, 'Lorem');
+    assert.deepEqual(page.foo().findBy('isSpecial').text, 'Lorem');
+
     let textArray = [];
     page.foo().forEach((i) => {
       textArray.push(i.text);


pzuraq:
The one thing I think may make sense to add is an assertion that the array element that is found is the _only_ element that matches the given value. Since `find` just returns the first value that it matches, tests could become brittle if developers rely on that behavior and it changes under their feet. If they do actually want to match multiple elements, they could continue to use `filterBy`

pzuraq:
related #288

addon/-private/properties/collection.js
```diff
@@ -5,7 +5,7 @@ import { create } from '../create';
 import { count } from './count';
 import Ceibo from 'ceibo';
 
-const arrayDelegateMethods = ['map', 'filter', 'mapBy', 'filterBy', 'forEach'];
+const arrayDelegateMethods = ['map', 'filter', 'find', 'mapBy', 'filterBy', 'findBy', 'forEach'];
```
magistrula: Will this work if the host app has disabled Array prototype extensions?

pzuraq:
`find` is a native function so it will continue to work. I’m unsure about `findBy` as it is an extension, but so is `filterBy` so I was following that pattern.

addon/-private/properties/collection.js
magistrula: It looks like we're not depending on the host app's prototype extensions and are ensuring that these functions are executed on an Ember Array: https://github.com/san650/ember-cli-page-object/blob/master/addon/-private/properties/collection.js#L23

magistrula:
@pzuraq I think something like `page.rows('text', 'foobar');` is a really interesting idea. What would you think about having it take this shape, so that you could query items based on multiple properties (similar to lodash `_.find`)?

```javascript
page.rows({ text: 'foobar', isVisible: true });
```

The next thing I wonder is, would it be natural for a developer to expect that syntax to return just one item as opposed to all matching items? Maybe it would best to have more clearly named custom `find` methods:

```javascript
// returns one item & throws an error if more than 1 is found
page.rows.findOne({ text: 'foobar', isVisible: true });

// returns all matching items
page.rows.findAll({ text: 'foobar', isVisible: true });
```

If we think it's appropriate, we could set up `findOne` and `findAll` to accept either an object or a function to make them more flexible. `findAll` would end up overlapping with `filter` if it took a function...dunno what I think about that yet.

Thoughts?

pzuraq:
I like the idea of matching the lodash find signature, it seems pretty flexible and a good compromise between terseness and expressiveness. I think being able to pass an object with key/values and a function to match against would be ideal, the other two syntaxes (`_.find(users, ['active', false]);` and `_.find(users, 'active');`) seem a little redundant, but we could include them too.

I agree, I'm not sure which result would be the most natural to expect. I think `findOne` and `findAll` would be great, and while `findAll` would have overlap with `filter` it would be much more powerful in being able to match against multiple values, so I think it would still be useful to add.

I think at this point we're starting to see the issues with the current syntax as well, where we must always invoke the function to get an enumerable first. I think we could change this in a backwards compat way by adding these new functions, along with others, to the prototype of the function:

```js
page.rows.eq(0); // query it like in the old style
page.rows.map(fn); // map over the collection
page.rows.forEach(fn);
```

We could start experimenting with a new syntax like this and deprecate the old style for v2 (or v3, if v2 is coming too soon)

pzuraq:
I'm going to spike that approach out tomorrow and open up a separate PR to compare the two approaches

san650:
Closing this for now. Feel free to open the PR again.

PR #346, opened 2017-11-19
Use project's jquery if available

ro0gr:
This should fix issues like the one described by @magistrula in slack:

> I ran my tests without calling `useNativeEvents`, and here’s what I’m observing: if I run my tests on `v.1.12.0`, some tests pass that expect an empty `div` not to be visible. If I run my tests on `v1.13.0-alpha.1`, those tests fail. However, this test fails in ember-cli-page-object if I checkout to `v1.12.0` and when I checkout to `v1.13.0-alpha.1`

Also it simplifies inclusion of `jquery`. Thanks to @pzuraq for the idea.

### DIFF
--- a/vendor/shims/ecpo-jquery.js
+++ b/vendor/shims/ecpo-jquery.js
@@ -1,12 +1,21 @@
-// Define an amd '-jquery' shim which is used by ember-cli-page-object internally
+// Define an amd module called "-jquery" 
+// which exposes jQuery bundeled with ember-cli-page-object.
+//
+// This mode is used when we deal with `jquery`-less apps.
 (function() {
-  var jquery = window.__ecpoJQuery__;
-  delete window.__ecpoJQuery__;
+  // prevent jquery provided by ember-cli-page-object
+  // to be set to `Ember.$` by Ember. 
+  var jq = self['$'].noConflict();
+  delete self['jQuery'];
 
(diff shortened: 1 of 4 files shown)


vendor/shims/ecpo-jquery.js
```diff
@@ -1,12 +1,21 @@
-// Define an amd '-jquery' shim which is used by ember-cli-page-object internally
+// Define an amd module called "-jquery" 
+// which exposes jQuery bundeled with ember-cli-page-object.
+//
+// This mode is used when we deal with `jquery`-less apps.
```
magistrula: Could we include in this comment something about the relationship between this shim and `vendor/ecpo-jquery/dist/jquery.js`?

ro0gr: > vendor/ecpo-jquery/dist/jquery.js

Do you want this path to be specified here? I'm not opposed but I think it can easily become outdated if we change the path in the `index.js`. 
I'm open for suggestions.

magistrula: Oh yah, no need to specify the path...just would be helpful to add some context about this shim is used. For example, it expects that we have already imported our own version of jQuery, right?

ro0gr: ok, seems reasonable

PR #347, opened 2017-11-20
feat(collection): Exposes a new Collection API 

pzuraq:
This PR implements the new Collection API RFC outlined in #352. The new API is optionally applied if users omit `itemScope` from collection definitions, allowing them to opt into the new syntax over time.

## Structure

The new collection is implemented as a class, that has a single instance per collection, rather than returning a new one each time the collection is called. Users are meant to interact with this class directly rather than calling the collection like method as before.

The class also caches individual item pages internally so they don't have to be regenerated each time the collection is accessed. This allows us to optimize for CPU over memory usage, which will (incrementally) help test speed.

## API

The new API that is exposed has the following methods:

* `objectAt`: Returns the item at the specified index, throws an error if it doesn't exist
* `length`: The current length of the collection
* `toArray`: Returns an array of all the items at that point in time. Should likely be a native array.
* `forEach`: Runs a function for every item in the collection, like native `forEach`
* `filter`: Filters the collection, like native `filter`
* `filterBy`: Filters the collection by a given key
* `map`: Maps over the collection, like native `map`
* `mapBy`: Maps over the collection by the given key

cc @magistrula @san650 @ro0gr

### DIFF
--- a/addon/-private/properties/collection.js
+++ b/addon/-private/properties/collection.js
@@ -1,113 +1,23 @@
-/* global Symbol */
-import Ember from 'ember';
-import { buildSelector, assign as mergeFunction } from '../helpers';
-import { create } from '../create';
-import { count } from './count';
-import Ceibo from 'ceibo';
+import { deprecate } from '@ember/application/deprecations';
 
-const arrayDelegateMethods = ['map', 'filter', 'mapBy', 'filterBy', 'forEach'];
-
-function merge(target, ...objects) {
-  objects.forEach((o) => mergeFunction(target, o));
-
-  return target;
-}
-
-function generateEnumerable(node, definition, item, key) {
-  let enumerable = merge({}, definition);
-
-  if (typeof (enumerable.count) === 'undefined') {
-    enumerable.count = count(item.itemScope);
-  }
-
-  if (typeof (enumerable.toArray) === 'undefined') {
-    enumerable.toArray = toArrayMethod(node, item, key);
-    arrayDelegateMethods.forEach((method) => delegateToArray(enumerable, method));
-  }
-
-  let collection = create(enumerable, { parent: node });
-
-  if (typeof (Symbol) !== 'undefined' && Symbol.iterator) {
-    collection[Symbol.iterator] = iteratorMethod;
-  }
-
-  // Change the key of the root node
-  Ceibo.meta(collection).key = `${key}()`;
-
-  return collection;
-}
-
-function generateItem(node, index, definition, key) {
-  let filters = merge({}, { scope: definition.scope, at: index });
-  let scope = buildSelector({}, definition.itemScope, filters);
-
-  let tree = create(
-    merge(
-      {
-        testContainer: definition.testContainer
-      },
-      definition.item,
-      {
-        scope,
-        resetScope: definition.resetScope
-      }
-    ), { parent: node });
-
-  // Change the key of the root node
-  Ceibo.meta(tree).key = `${key}(${index})`;
-
-  return tree;
-}
-
-function toArrayMethod(node, definition, key) {
-  return function() {
-    let array = Ember.A();
-    let index;
-    let count;
-
-    for (index = 0, count = this.count; index < count; index++) {
-      array.push(generateItem(node, index, definition, key));
-    }
-
-    return array;
-  };
-}
-
-function delegateToArray(enumerable, method) {
-  if (typeof (enumerable[method]) === 'undefined') {
-    enumerable[method] = function(...args) {
-      return this.toArray()[method](...args);
-    };
-  }
-}
-
-function iteratorMethod() {
-  let i = 0;
-  let items = this.toArray();
-  let next = () => ({ done: i >= items.length, value: items[i++] });
-
-  return { next };
-}
+import { collection as mainCollection } from './collection/main';
+import { collection as legacyCollection } from './collection/legacy';
 
 /**
  * @public
  *
- * Creates a component that represents a collection of items. The collection is zero-indexed.
+ * Creates a enumerable that represents a collection of items. The collection is zero-indexed
+ * and has the following public methods and properties:
  *
- * When called with an index, the method returns the matching item.
- *
- * When called without an index, the the object returned behaves as a regular PageObject with a few additional properties and methods:
- *
- * - `count` - the number of items in the collection
+ * - `length` - The number of items in the collection.
+ * - `objectAt()` - Returns the page for the item at the specified index.
+ * - `filter()` - Filters the items in the array and returns the ones which match the predicate function.
+ * - `filterBy()` - Filters the items of the array by the specified property, returning all that are truthy or that match an optional value.
+ * - `forEach()` - Runs a function for each item in the collection
+ * - `map()` - maps over the elements of the collection
+ * - `mapBy()` - maps over the elements of the collecton by the specified property
  * - `toArray()` - returns an array containing all the items in the collection
  * - `[Symbol.iterator]()` - if supported by the environment, this allows the collection to be iterated with `for/of` and spread with `...` like a normal array
- *
- * Collection objects also delegate the following methods to `toArray()` for ease of consumption:
- * - `map`
- * - `mapBy`
- * - `filter`
- * - `filterBy`
- * - `forEach`
  *
  * @example
  *
@@ -127,21 +37,15 @@ function iteratorMethod() {
  *
  * const page = PageObject.create({
  *   users: collection({
- *     itemScope: 'table tr',
- *
- *     item: {
- *       firstName: text('td', { at: 0 }),
- *       lastName: text('td', { at: 1 })
- *     },
- *
- *     caption: text('caption')
+ *     scope: 'table tr',
+ *     firstName: text('td', { at: 0 }),
+ *     lastName: text('td', { at: 1 })
  *   })
  * });
  *
- * assert.equal(page.users().count, 2);
- * assert.equal(page.users().caption, 'List of users');
- * assert.equal(page.users(1).firstName, 'John');
- * assert.equal(page.users(1).lastName, 'Doe');
+ * assert.equal(page.users.length, 2);
+ * assert.equal(page.users.objectAt(1).firstName, 'John');
+ * assert.equal(page.users.objectAt(1).lastName, 'Doe');
  *
  * @example
  *
@@ -166,19 +70,16 @@ function iteratorMethod() {
  * // </div>
  *
  * const page = PageObject.create({
+ *   scope: '.admins',
+ *
  *   users: collection({
- *     scope: '.admins',
- *
- *     itemScope: 'table tr',
- *
- *     item: {
- *       firstName: text('td', { at: 0 }),
- *       lastName: text('td', { at: 1 })
- *     }
+ *     scope: 'table tr',
+ *     firstName: text('td', { at: 0 }),
+ *     lastName: text('td', { at: 1 })
  *   })
  * });
  *
- * assert.equal(page.users().count, 2);
+ * assert.equal(page.users.length, 2);
  *
  * @example
  *
@@ -186,58 +87,41 @@ function iteratorMethod() {
  * //   <caption>User Index</caption>
  * //   <tbody>
  * //     <tr>
- * //       <td>Doe</td>
- * //     </tr>
+ * //         <td>Mary<td>
+ * //         <td>Watson</td>
+ * //       </tr>
+ * //       <tr>
+ * //         <td>John<td>
+ * //         <td>Doe</td>
+ * //       </tr>
  * //   </tbody>
  * // </table>
  *
  * const page = PageObject.create({
+ *   scope: 'table',
+ *
  *   users: PageObject.collection({
- *     scope: 'table',
- *     itemScope: 'tr',
- *
- *     item: {
- *       firstName: text('td', { at: 0 })
- *     },
- *
- *     caption: PageObject.text('caption')
+ *     scope: 'tr',
+ *     firstName: text('td', { at: 0 }),
+ *     lastName: text('td', { at: 1 }),
  *   })
  * });
  *
- * assert.equal(page.users().caption, 'User Index');
+ * let john = page.users.filter((item) => item.firstName === 'John' )[0];
+ * assert.equal(john.lastName, 'Doe');
  *
  * @param {Object} definition - Collection definition
- * @param {string} definition.scope - Nests provided scope within parent's scope
- * @param {boolean} definition.resetScope - Override parent's scope
- * @param {string} definition.itemScope - CSS selector
- * @param {Object} definition.item - Item definition
  * @return {Descriptor}
  */
 export function collection(definition) {
-  definition = mergeFunction({}, definition);
+  if ('itemScope' in definition) {
+    deprecate('You are currently using the legacy collection API, check the documentation to see how to upgrade to the new API.', false, {
+      id: 'ember-cli-page-object.old-collection-api',
+      until: '2.0.0'
+    });
 
-  let item = {
-    scope: definition.scope,
-    itemScope: definition.itemScope,
-    resetScope: definition.resetScope,
-    item: definition.item,
-    testContainer: definition.testContainer
-  };
-
-  delete definition.item;
-  delete definition.itemScope;
-
-  return {
-    isDescriptor: true,
-
-    get(key) {
-      return (index) => {
-        if (typeof (index) === 'number') {
-          return generateItem(this, index, item, key);
-        } else {
-          return generateEnumerable(this, definition, item, key);
-        }
-      };
-    }
-  };
+    return legacyCollection(definition);
+  } else {
+    return mainCollection(definition);
+  }
 }
--- a/addon/-private/properties/collection/legacy.js
+++ b/addon/-private/properties/collection/legacy.js
@@ -1,0 +1,119 @@
+/* global Symbol */
+import Ember from 'ember';
+import { buildSelector, assign as mergeFunction } from '../../helpers';
+import { create } from '../../create';
+import { count } from '../count';
+import Ceibo from 'ceibo';
+
+const arrayDelegateMethods = ['map', 'filter', 'mapBy', 'filterBy', 'forEach'];
+
+function merge(target, ...objects) {
+  objects.forEach((o) => mergeFunction(target, o));
+
+  return target;
+}
+
+function generateEnumerable(node, definition, item, key) {
+  let enumerable = merge({}, definition);
+
+  if (typeof (enumerable.count) === 'undefined') {
+    enumerable.count = count(item.itemScope);
+  }
+
+  if (typeof (enumerable.toArray) === 'undefined') {
+    enumerable.toArray = toArrayMethod(node, item, key);
+    arrayDelegateMethods.forEach((method) => delegateToArray(enumerable, method));
+  }
+
+  let collection = create(enumerable, { parent: node });
+
+  if (typeof (Symbol) !== 'undefined' && Symbol.iterator) {
+    collection[Symbol.iterator] = iteratorMethod;
+  }
+
+  // Change the key of the root node
+  Ceibo.meta(collection).key = `${key}()`;
+
+  return collection;
+}
+
+function generateItem(node, index, definition, key) {
+  let filters = merge({}, { scope: definition.scope, at: index });
+  let scope = buildSelector({}, definition.itemScope, filters);
+
+  let tree = create(
+    merge(
+      {
+        testContainer: definition.testContainer
+      },
+      definition.item,
+      {
+        scope,
+        resetScope: definition.resetScope
+      }
+    ), { parent: node });
+
+  // Change the key of the root node
+  Ceibo.meta(tree).key = `${key}(${index})`;
+
+  return tree;
+}
+
+function toArrayMethod(node, definition, key) {
+  return function() {
+    let array = Ember.A();
+    let index;
+    let count;
+
+    for (index = 0, count = this.count; index < count; index++) {
+      array.push(generateItem(node, index, definition, key));
+    }
+
+    return array;
+  };
+}
+
+function delegateToArray(enumerable, method) {
+  if (typeof (enumerable[method]) === 'undefined') {
+    enumerable[method] = function(...args) {
+      return this.toArray()[method](...args);
+    };
+  }
+}
+
+function iteratorMethod() {
+  let i = 0;
+  let items = this.toArray();
+  let next = () => ({ done: i >= items.length, value: items[i++] });
+
+  return { next };
+}
+
+export function collection(definition) {
+  definition = mergeFunction({}, definition);
+
+  let item = {
+    scope: definition.scope,
+    itemScope: definition.itemScope,
+    resetScope: definition.resetScope,
+    item: definition.item,
+    testContainer: definition.testContainer
+  };
+
+  delete definition.item;
+  delete definition.itemScope;
+
+  return {
+    isDescriptor: true,
+
+    get(key) {
+      return (index) => {
+        if (typeof (index) === 'number') {
+          return generateItem(this, index, item, key);
+        } else {
+          return generateEnumerable(this, definition, item, key);
+        }
+      };
+    }
+  };
+}
--- a/addon/-private/properties/collection/main.js
+++ b/addon/-private/properties/collection/main.js
@@ -2,6 +2,10 @@
 import Ember from 'ember';
 import { buildSelector, assign } from '../../helpers';
 import { create } from '../../create';
+import {
+  ELEMENT_NOT_FOUND,
+  throwBetterError
+} from '../../better-errors';
 import { count } from '../count';
 import Ceibo from 'ceibo';
 
@@ -42,7 +46,18 @@ export class Collection {
       this._items[index] = tree;
     }
 
-    return this._items[index];
+    let itemPage = this._items[index];
+
+    if (index > this.length - 1) {
+      throwBetterError(
+        itemPage,
+        null,
+        ELEMENT_NOT_FOUND,
+        { selector: itemPage.scope }
+      );
+    }
+
+    return itemPage;
   }
 
   filter(...args) {
--- a/addon/-private/properties/collection/main.js
+++ b/addon/-private/properties/collection/main.js
@@ -1,0 +1,104 @@
+/* global Symbol */
+import Ember from 'ember';
+import { buildSelector, assign } from '../../helpers';
+import { create } from '../../create';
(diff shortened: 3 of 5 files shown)


addon/-private/properties/collection.js
```diff
@@ -1,92 +1,177 @@
...
+
+  findAll(query) {
+    let predicate;
+
+    if (typeof (query) === 'object') {
+      predicate = (item) => {
+        let isMatch = true;
+
+        for (let key in query) {
+          isMatch = isMatch && item[key] === query[key];
```
magistrula: I wonder if we need to support array comparisons here. 

Suppose some parking lot app allows attendants to enter a list of driver names, each of which is followed by a series of inputs used to enter information about the driver's car:
```html
<div data-test-driver>
  <input type="text" placeholder="Driver Name">
  <input type="text" placeholder="Car Make">
  <input type="text" placeholder="Car Model">
  <input type="text" placeholder="Car License Plate">
</div>
```
```javascript
// in the page object . . .
drivers: collection({
  scope: '[data-test-driver]',
  driverInfo: text('input', { multiple: true })
})
```
```javascript
// in some test . . .
const driver = page.drivers.findOne({ 
  driverInfo: ['Grace Hopper', 'Nissan', 'Sentra', 'FOOBAR'] 
});
```

Now, this would make a lot more sense:

```html
<div data-test-driver>
  <input type="text" data-test-driver-name placeholder="Driver Name">
  <input type="text" data-test-car-make placeholder="Car Make">
  <input type="text" data-test-car-model placeholder="Car Model">
  <input type="text" data-test-car-plate placeholder="Car License Plate">
</div>
```
```javascript
// in the page object . . .
drivers: collection({
  scope: '[data-test-driver]',
  driverName: text('[data-test-driver-name]'),
  carMake: text('[data-test-car-make]'),
  carModel: text('[data-test-car-model]'),
  carPlate: text('[data-test-car-plate]')
})
```
```javascript
// in some test . . .
const driver = page.drivers.findOne({ 
  driverName: 'Grace Hopper',
  carMake: 'Nissan', 
  carModel: 'Sentra', 
  carPlate: 'FOOBAR'
});
```
Even though the first version up above is not what we'd recommend, someone could nevertheless set up their page that way, and it might be confusing that `findOne({ driverInfo: ['Grace Hopper', 'Nissan', 'Sentra', 'FOOBAR'] })` doesn't work.

Thoughts?

addon/-private/properties/collection.js
```diff
@@ -1,92 +1,177 @@
...
+        }
+
+        return isMatch;
+      }
+    } else if (typeof (query) === 'function') {
+      predicate = query;
+    } else {
+      assert(`Expected query for findAll to be either an object or function, received: ${query}`, false);
```
magistrula: Can we use `throwBetterError` here?

addon/-private/properties/collection.js
```diff
@@ -1,92 +1,177 @@
...
+    return this.filter(predicate);
   }
 }
 
-function iteratorMethod() {
-  let i = 0;
-  let items = this.toArray();
-  let next = () => ({ done: i >= items.length, value: items[i++] });
+if (typeof (Symbol) !== 'undefined' && Symbol.iterator) {
+  Collection.prototype[Symbol.iterator] = function() {
```
magistrula: Ah, cool!

addon/-private/properties/collection.js
pzuraq: I think at that point we would have to make the comparison a `deepEqual`, it would be odd to support a single level of nested comparison. I was mostly following the simple example form `lodash`, I'm unsure if they support deep equals in [find and filter](https://lodash.com/docs/4.17.4#find), but I definitely could implement it without too much trouble.

I think that this could allow users to do some interesting and maybe counter-intuitive things, so I'm split on whether it would be a good thing to add. I think I would be more comfortable with keeping it as is, without deep-comparison, because that is the more minimal API: If we add deep comparison now, we can't remove it later without making breaking changes, but we can always add it if we omit it for now.

We also have the escape hatch of allowing users to pass the predicate themselves, so they could do something like:

```js
const driver = page.drivers.findOne(driver => deepEqual(driver, { 
  driverInfo: ['Grace Hopper', 'Nissan', 'Sentra', 'FOOBAR'] 
}));
```

magistrula: As I was sanity-checking this line, the array example is the only real-world situation I could come up with where strict equality comparison wouldn't work, but I agree...if we were to do anything besides strict comparison, we would have to go with full-on `deepEqual`. 

I was mainly thinking here about predictability. Lodash `_.find` does allow something like `_.find(page.drivers.toArray(), { driverInfo: ['Grace Hopper', 'Nissan', 'Sentra', 'FOOBAR'] })`. Underscore, however, does not. `Ember.Array.findBy` does not. So...fair enough to go in either direction with this particular bit of syntactic sugar. I like your point about how we can always add this later. If there is a real need for deeper equality comparison, someone will raise it as an issue. I'm down with moving forward with strict comparison for now.

addon/-private/properties/collection/main.js
```diff
@@ -0,0 +1,137 @@
...
+
+export class Collection {
+  constructor(definition, parent, key) {
+    this.definition = definition;
+    this.parent = parent;
+    this.key = key;
+
+    this._countPage = create({
+      count: count(definition.scope, { testContainer: definition.testContainer })
+    }, { parent });
```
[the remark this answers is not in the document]
pzuraq: Ah, yes, we definitely do! I'll update it, and I agree that we can use a better name. I like `_itemCounter` instead, it reads cleaner

addon/-private/properties/collection/main.js
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  objectAt(index) {
+    if (this._items[index] === undefined) {
+      let { definition, parent, key } = this;
+      let scope = buildSelector({}, definition.scope, { at: index });
+
+      let finalizedDefinition = assign({}, definition);
+
+      finalizedDefinition.scope = scope;
```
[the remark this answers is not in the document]
pzuraq: This is an old habit, `Ember.merge` couldn't assign arbitrary numbers of objects, but `Ember.assign` and `Object.assign` can. 

It looks like `assign` is actually either `Ember.assign || Ember.merge`, which could be pretty dangerous here. I'm assuming that's for backwards compatibility. It also explains why there's a recursive merge function in the old collection code: https://github.com/san650/ember-cli-page-object/blob/master/addon/-private/properties/collection.js#L10

I'm open to changing the root merge function instead, it seems like it should either be `Ember.assign` or a polyfill.

addon/-private/properties/collection/main.js
```diff
@@ -0,0 +1,137 @@
...
+        this.parent,
+        this.key,
+        `Expected at most one result from findOne query, but found ${result.length}`
+      );
+    }
+
+    return result[0];
+  }
+
+  findAll(query) {
```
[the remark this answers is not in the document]
pzuraq: Yes, exactly. I debated including `filter` and `filterBy`, but I decided against it because they are pretty redundant, but less safe (no assertions, etc) and not the recommended path. On top of that, if we included them it wouldn't make much sense to exclude other array functions, like `find` (which would be doubly confusing) and others.

addon/-private/properties/collection/main.js
```diff
@@ -0,0 +1,137 @@
...
+    isDescriptor: true,
+
+    setup(node, key) {
+      // Set the value on the descriptor so that it will be picked up and applied by Ceibo.
+      // This does mutate the descriptor, but because `setup` is always called before the
+      // value is assigned we are guaranteed to get a new, unique Collection instance each time.
+      descriptor.value = new Collection(this._definition, node, key);
+    },
+
+    _definition: definition
```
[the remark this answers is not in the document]
pzuraq: Ah, so this is actually completely for the benefit of `ember-classy-page-object` I just realized, so I would definitely be open to changing it or removing it.

The key issue here is that because collections are a function and return a descriptor, there is no way to access the definition externally. So, you can't extend a collection's definition later on. Basically, there's no way to do:

```js

let pageOne = {
  items: collection({
    scope: 'li'
  })
}

let pageTwo = {
  items: collection({
    ...pageOne.items,

    click: clickable()
  })
}
```

This would make that definition accessible later on. There are workarounds for this that I can use in `ember-classy-page-object`, though they aren't super pretty. Either way, I'm open.

addon/-private/properties/collection/main.js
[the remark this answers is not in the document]
magistrula: Yeah...feels tricky having to reason about which `assign` behavior we're getting here. Let's leave this as is for now. @bendemboski has a [PR](#357) open that will change the `assign` function provided by `helpers.js`, so we can come back to this later.

addon/-private/properties/collection/main.js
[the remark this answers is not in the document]
magistrula: Yah, I think I'd prefer to remove for now and tackle extending collections as a separate issue, just to reduce the problems to solve & reason about in this PR. It's an interesting issue that probably deserves its own discussion.

addon/-private/properties/collection/main.js
```diff
@@ -0,0 +1,140 @@
...
+
+  mapBy(...args) {
+    return this.toArray().mapBy(...args);
+  }
+
+  forEach(...args) {
+    return this.toArray().forEach(...args);
+  }
+
+  findOne(query) {
```
ro0gr: Any reasons not to call it `find()`?

addon/-private/properties/collection/main.js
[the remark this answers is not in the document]
ro0gr: I'd prefer `filter()` more for the sake of learnability.
How does `findAll` differ from `filter`? It seems they both have the same signatures and do very similar thing.

addon/-private/properties/collection/main.js
pzuraq: `find` is a native array method and only accepts a function. `findOne` can receive a general query object, matching the signature of lodash's `find` method. It also asserts that _exactly_ one match exists, whereas `find` will find the _first_ instance, but also allows other instances to exist in the array. The difference in behavior is big enough we decided to rename the function to avoid confusion.

addon/-private/properties/collection/main.js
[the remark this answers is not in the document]
pzuraq: `findAll` can accept an object that is a query, matching lodash's `filter/find` methods. We decided not to name it `filter` since A. it may be confusing to users who expect it to work exactly like native array's `filter`, and B. we may want to add `filter/filterBy` passthroughs at some point in the future.

addon/-private/properties/collection/main.js
ro0gr: > It also asserts that exactly one match exists

I believe this should be a responsibility of the test framework:
```js
  const visibleItems = page.items.findAll(i => i.isVisible);
  assert.equal(visibleItems.length, 1); // the user also can specify assertion message specific to the testing context
```

ro0gr: > findOne can receive a general query object

I think understand the reason behind adding a declarative query objects feature. However I think it's quite complicated area to be done properly. I mean without any specific DSL it has only strict equals. We can't use `less then`, `greater then`, `contains`. Also it's not obvious how to deal with nested objects:
```js
page.items.findOne({
  isVisible: true,
  input: {
    value: 'test' // is this kind of comparison supported?
  }
})
``` 

And the last, by comparing function vs query object I can't see big benefits in using the former:
```js
page.items.findAll(i => i.isVisible)
page.items.findAll({ isVisible: true })
```

pzuraq: A better example for the benefits of the query object would be with string values, or multiple values:

```
page.items.findAll(item => item.name === 'S&P 500 Stock Index' && item.isVisible);

page.items.findAll({ 
  name: 'S&P 500 Stock Index',
  isVisible: true
});
```

Really this is just a way to provide shorthand. The current API proposed in `findAll` and `findOne` was proposed in #345, originally I suggested adding the `findBy` method but @magistrula suggested using query objects instead since they are more flexible.

As for the assertion about only one result existing, I personally disagree. Relying on implicit behavior is the reason tests become brittle - for instance, relying on index rather than content of an item in a collection. If a user thinks they are selecting the _only_ instance of an item, but they are really selecting the _first_ instance, the resulting tests only have more chance of being brittle and breaking due to unrelated changes, especially in a way that is hard to debug. However, if we enforce this assertion then most users will get the behavior they want (find the one item they were looking for) and the rest can default to using `findAll` instead.

I'm also open to switching back to `filter` and `filterBy` and implementing `find` and `findBy` as well, but I think this is an opportunity to add more layers of safety that help prevent users from making brittle tests.

pzuraq: It's also worth noting, we can do either strategy right now and implement the other later on. So we could stick to native array methods for now and consider adding `findOne` and `findAll` later on, or vice-versa. I'd definitely prefer that if we can get this merged since the other benefits are pretty massive.

ro0gr: I'm not sure about the right path forward. Maybe @magistrula @san650 can help here?

There are several interesting features like query objects and `findOne` but I feel like they are out of scope here. I'd like to ship new collections in the slimmest possible form and iterate on new features in a separate PRs.

ro0gr: If I cheat a little bit it I can't see benefits again 😆 :
```js
page.items.findAll(i => 
  i.name === 'S&P 500 Stock Index'
  && i.isVisible
);

page.items.findAll({ 
  name: 'S&P 500 Stock Index',
  isVisible: true
});
```

> since they are more flexible

I don't think object queries are flexible in its current form. I've tried to explain why [here](using query objects instead since they are more flexible).

Actually I would rather deprecate `filterBy` cause it seems like a weak api.

---

Regarding `findOne` you made me doubt @pzuraq. It seems more reasonable to me now.

magistrula: Seems like there's enough discussion of `findOne` and `findAll` here that I think it would make sense to simplify this PR by pulling those into a separate PR, but will that mean that we need to decide whether to provide `filter` and `filterBy` in this PR? Or can we leave those out as well and decide about them when we discuss `findOne` and `findAll`? Do we think that `filter`, `filterBy`, `findOne`, and `findAll` could all live alongside each other in the new API?

pzuraq:
I think we should provide `filter` and `filterBy` and that they can live alongside the other methods in the future. Do you think we should include `find` and `findBy` as well to make us consistent with array/Ember.A apis?

magistrula:
I feel like `find`, `findBy`, and `findOne` could live alongside each other, but I'm not so sure about adding `findAll` to that group. The only thing `findAll` offers over `filter` is that you can pass a query object to it, and it feels like it would be weird to have two methods that are basically the same except that one is a little more sugary than the other. I really like the convenience of passing a query object to `findOne`, but it muddies the waters when it comes to `filter` vs `findAll`. What do you think about providing `find`, `findBy`, `findOne` (all of which would take a function) and dropping the idea of a query object and `findAll`?

ro0gr:
I think we should not provide both `find` and `findOne` cause `findOne` effectively makes `find` a redundant method. 
If that's true I also like an idea not to include `findBy` to the API. The reason behind that is that it's very limited and doesn't provide any value I believe:
```js
  page.items.findBy('isVisible', true);
  // vs
  page.items.find(i => i.isVisible);
```

ro0gr:
Also regarding `findOne` vs `find` name. As far as I can see the real issue why we don't want to call it `find` is a new guard for a single selected item. 
However I'm pretty sure the user never wants a `find` behavior without this guard. This guard can be considered as an smart assumption of `ember-cli-page-object`'s `find` method. Of course this guard should be explicitly explained in the docs.
If that sounds reasonable we could have nice and predictable API names for collection I believe.

pzuraq:
Updated again to remove `findAll` and `findOne`, we can add the new `find` functions in a future PR. I think this is ready for review again.

I can't seem to figure out what's causing coveralls to fail. There don't seem to be any files in the report that have less coverage, but it also looks like not all the files are in the report. Can anyone help me figure out what's going on there?

addon/-private/properties/collection/legacy.js
```diff
@@ -0,0 +1,119 @@
+/* global Symbol */
```
ro0gr: can be moved to `.eslintrc`

pzuraq:
We have, there is `collection-test.js` and `legacy-collection-test.js` which tests the main and legacy versions respectively

addon/-private/properties/collection/main.js
```diff
@@ -0,0 +1,104 @@
...
+    let { length } = this;
+
+    let array = Ember.A();
+
+    for (let i = 0; i < length; i++) {
+      array.push(this.objectAt(i));
+    }
+
+    return array;
+  }
```
magistrula: Just an idea...
```javascript
toArray() {
  return Ember.A(Array.from(this));
}

...

if (typeof (Symbol) !== 'undefined' && Symbol.iterator) {
  Collection.prototype[Symbol.iterator] = function() {
    let i = 0;
    let next = () => ({ done: i >= this.length, value: this.objectAt(i++) });
    return { next };
  }
}
```

addon/-private/properties/collection/main.js
```diff
@@ -0,0 +1,104 @@
...
+
+    this._items = [];
+  }
+
+  get length() {
+    return this._itemCounter.count;
+  }
+
+  objectAt(index) {
+    if (typeof this._items[index] === 'undefined') {
```
magistrula: What do you think of having this return `undefined` if `index >= this.length`?

addon/-private/properties/collection/main.js
pzuraq: that would be cool, but then we would also be assume that `Symbol` and `Symbol.iterator` are supported no?

addon/-private/properties/collection/main.js
pzuraq: I like that idea, brings us inline with actual array semantics. Will update and add a test

tests/unit/-private/properties/collection-test.js
```diff
@@ -373,46 +199,29 @@ moduleForProperty('collection', function(test) {
 
     this.adapter.createTemplate(this, page);
 
-    assert.throws(() => page.foo.bar().baz.qux, /page\.foo\.bar\(\)\.baz\.qux/);
-  });
-
-  test("doesn't generate an item or itemScope property", function(assert) {
```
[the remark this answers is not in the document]
pzuraq: Actually this is checking that the enumerable itself doesn't get those properties, only the individual page objects:

```
assert.notOk(page.foo().item);		
assert.notOk(page.foo().itemScope);

assert.ok(page.foo(0).item);		
assert.ok(page.foo(0).itemScope);
```

Moreso a weird quirk of the original collection having that wrapper page object. We can test 0 length collections though

addon/-private/properties/collection/main.js
pzuraq: Actually, this does interfere with the current behavior which is based on the legacy collection's behavior, see [this test](https://github.com/pzuraq/ember-cli-page-object/blob/5f2bb1d67802faf94809d7b23071ef4ded535830/tests/unit/-private/properties/collection-test.js#L189-L203).

I'm open to either behavior being correct, we could also throw an error when they first try to get the page itself.

magistrula: Ah, yeah, I guess throwing an error makes the most sense, since that's what other page object methods do if you try to interact with an element that doesn't exist.

addon/-private/properties/collection/main.js
magistrula: Ah, yeah...good point.

addon/-private/properties/collection/main.js
pzuraq: Returning to this, I think an error may actually be too harsh of an assertion. I think your first instinct was right, returning `undefined` brings us inline with standard `objectAt` behavior

pzuraq: Specifically, asseting prevents users from checking whether an item exists or not, and makes that case always an error, which I think could be really counterintuitive. Like if you just deleted an item and wanted to check if it was gone

magistrula: On the other hand, if you wanted to check if it was gone, you could check `length`. I guess if we want to be consistent with the expectations that `objectAt` sets up for developers, we should return `undefined`. If we want to be consistent with other page object helpers, we should throw an error. Does that seem like an accurate description of the decision we're facing? @san650, do you have a preference for what happens when you try to access an item in a collection that does not exist? Do you have a preference for whether we get `undefined` or an element-not-found error?

san650: @magistrula @pzuraq I think it should throw an element-not-found error. When you're writing tests, your asserting on things you know they should be true, so for me its weird to receive an `undefined`.

tests/unit/-private/properties/legacy-collection-test.js
```diff
@@ -0,0 +1,584 @@
...
+      page,
+      '<span>Lorem</span><span>ipsum</span>',
+      { useAlternateContainer: true }
+    );
+
+    assert.equal(page.foo().count, 2);
+    assert.equal(page.foo(0).text, 'Lorem');
+  });
+
+  test('can provide custom array methods', function(assert) {
```
ro0gr: I'm afraid even to think why this functionality is needed 😨

magistrula: Yeah, that's a great point...maybe we should not allow this.

pzuraq: This is for the legacy collection only, so it is providing existing behavior. The reason I added the test was to get code coverage passing. 

Part of the failure was because of [this if statement here](https://github.com/pzuraq/ember-cli-page-object/blob/1402a5e69adc37d7c92b7b25ff360a2a6cbd517f/addon/-private/properties/collection/legacy.js#L30-L32). Because it is dependent on the window/browser, we cannot test the `else` case (that `Symbol` is undefined). Unfortunately, we had to add another similar `if` statement in the new implementation, which was also untestable, lowering code coverage overall. The only way I found around this was to add tests for previous behavior that was not tested, but did exist.

magistrula: Ah, right. The code coverage issue. So when we pull out the old collection & its tests, we'll get another coverage failure?

addon/-private/properties/collection/main.js
ro0gr: I'm probably a little bit late to the party. Anyways..



When we try to access a regular page object property which doesn't exist in the DOM we don't throw any errors. Instead we return a page object node which allows `isPresent` and `isVisible` to be executed.

tests/unit/-private/properties/legacy-collection-test.js
pzuraq: I believe overall code coverage will actually go up at that point, because we'll be removing all the cases _and_ one of the `Symbol` if statements

addon/-private/properties/collection/main.js
ro0gr: > Specifically, asseting prevents users from checking whether an item exists or not, and makes that case always an error, which I think could be really counterintuitive

Agree

magistrula: In this case, it feels less to me like we're accessing a page object property that doesn't exist and more like we're accessing an element that doesn't exist. Thoughts?

pzuraq: I think it's a bit of a paradigm shift, because before page objects always existed. The items that they referred to may-or-may not be in the DOM, but that didn't affect whether or not the page for the item existed - which is why you would have to assert `isPresent` on the individual item rather than checking if the page exists.

To be fair, I think that could trip up new users as well. Currently if we do `items(123456789)` it will return a page, has that ever thrown a user off?

tests/unit/-private/properties/legacy-collection-test.js
ro0gr: Sad but true.. I've never liked this partially supported `Symbol.Iterator` feature 😆

addon/-private/properties/collection/main.js
ro0gr: > Currently if we do `items(123456789)` it will return a page, has that ever thrown a user off?

I think there no any usages in the wild of `items(123456789)` exactly in this form. The user always wants to have an access to some page object property from the item:
```
  items(123456789).isVisible
  items(123456789).fillIn('something') // I think here we can assert as we do now
  items(123456789).title.click() // that's why `undefined` is not the best option
```

I think we already have a well defined behavior for all this cases and I find it pretty intuitive.

tests/unit/-private/properties/collection-test.js
```diff
@@ -214,7 +214,7 @@ moduleForProperty('collection', function(test) {
 
     this.adapter.createTemplate(this, page);
 
-    assert.throws(() => page.foo.bar.objectAt(1).baz.qux, /page\.foo\.bar\[1\]\.baz\.qux/);
+    assert.throws(() => page.foo.bar.objectAt(1).baz.qux, /page\.foo\.bar\[1\]/);
```
magistrula: We just need to update the `test` text to say something like `'throws an error when the item's element doesn't exist'`.

addon/-private/properties/collection/main.js
```diff
@@ -0,0 +1,119 @@
...
+
+      // Change the key of the root node
+      Ceibo.meta(tree).key = `${key}[${index}]`;
+
+      this._items[index] = tree;
+    }
+
+    let itemPage = this._items[index];
+
+    if (index > this.length - 1) {
```
magistrula: What do you think about moving this up after line 35?
```javascript
throwBetterError(
  parent,
  `${key}[${index}]`,  // we could save this to a variable so as not to repeat it here and on line 44
  ELEMENT_NOT_FOUND,
  { selector: scope }
)
```

magistrula: Oh nah, I see why that won't work...would only work if an item hadn't yet been created for that index. But what do you think about changing the first two args as suggested above?

addon/-private/properties/collection/main.js
```diff
@@ -0,0 +1,119 @@
...
+      throwBetterError(
+        itemPage,
+        null,
```
magistrula: Currently, this gives us an error like this:
```
Error: Element not found.  
  PageObject: 'page.foo.bar[1].'   
  Selector: '[data-test-bar]:eq(1)'
```

But this might be a bit nicer:

```
Error: Element not found.
  PageObject: 'page.foo.objectAt(1)'
  Selector: '[data-test-bar]:eq(1)'
```

So suggest this:
```javascript
throwBetterError(
  this.parent,
  `objectAt(${index})`,
  ELEMENT_NOT_FOUND,
  { selector: itemPage.scope }
);
```

PR #348, opened 2017-12-13
Fastboot Safe

demersus:
Our fastboot builds were failing for local development.  This prevents the addon from enabling itself in the fastboot build. (Addon depends on jquery)

### DIFF
--- a/index.js
+++ b/index.js
@@ -50,7 +50,7 @@ module.exports = {
     // https://github.com/joostdevries/twiddle-backend/pull/28 is merged.
     // return !!this.app.tests;
 
-    if('EMBER_CLI_FASTBOOT' in process.env) {
+    if(process.env && process.env.EMBER_CLI_FASTBOOT) {
       return false;
     } else {
       return this.app.env !== 'production';
--- a/index.js
+++ b/index.js
@@ -49,7 +49,12 @@ module.exports = {
     // TODO: In order to make the addon work in EmberTwiddle, we cannot use // the `tests` prop til
     // https://github.com/joostdevries/twiddle-backend/pull/28 is merged.
     // return !!this.app.tests;
-    return this.app.env !== 'production';
+
+    if('EMBER_CLI_FASTBOOT' in process.env) {
+      return false;
+    } else {
+      return this.app.env !== 'production';
+    }
   },
 
   _findHost() {


index.js
```diff
@@ -49,7 +49,12 @@ module.exports = {
     // TODO: In order to make the addon work in EmberTwiddle, we cannot use // the `tests` prop til
     // https://github.com/joostdevries/twiddle-backend/pull/28 is merged.
     // return !!this.app.tests;
-    return this.app.env !== 'production';
+
+    if('EMBER_CLI_FASTBOOT' in process.env) {
```
ro0gr: I think it would fail when:
```
EMBER_CLI_FASTBOOT=false ember s
```
can we change this condition to check if the value is true, something like `process.env.EMBER_CLI_FASTBOOT === 'true'`?

ro0gr:
@demersus I don't have experience with fastboot but I'm curious about `EMBER_CLI_FASTBOOT` variable. Do you set it manually or is it fastboot built-in feature?

san650:
Thanks @demersus for the PR.

As @ro0gr suggested, can you update the code to check that the value is `'true'`?

```js
process.env.EMBER_CLI_FASTBOOT === 'true';
```

@ro0gr doing a quick search in github, it seems that this VAR is what people uses to check this.

About the failings in the CI, they don't seem related with this PR, it seems that ember-data breaks our test suite 😞

demersus:
Yes, EMBER_CLI_FASTBOOT is present when this is building/running in fastboot.  I'll update the check to check for the correct value.  No idea what went wrong in the CI tests

demersus:
As for the `process.env.EMBER_CLI_FASTBOOT` treated as a boolean itself:  I've seen this pattern used in other projects on Github.  It seems to work in my test app fine.

ro0gr:
@san650 I've re-ran failing builds. The issue seems to be fixed now. Is this good enough to be merged?

san650:
Thanks @demersus for contributing!

bobisjan:
Hi,

I can tell that `EMBER_CLI_FASTBOOT` flag is not available in FastBoot 1.0+, https://github.com/ember-fastboot/ember-cli-fastboot/pull/369, https://github.com/ember-fastboot/ember-cli-fastboot/pull/399 and thus should not be used any more.

In our application we have both stable `ember-cli-fastboot` and `ember-cli-page-object` without any problems for some time.

I've tested `ember-cli-page-object@1.13.0-alpha.1` within our application and can confirm that it's failing to serve in FastBoot due to:

```
TypeError: Cannot read property 'noConflict' of undefined
    at /.../tmp/broccoli_persistent_filtersimple_replace-output_path-xKFvdG3g.tmp/assets/vendor/shims/ecpo-jquery-global.js:7:1
    at /.../tmp/broccoli_persistent_filtersimple_replace-output_path-xKFvdG3g.tmp/assets/vendor/shims/ecpo-jquery-global.js:9:1
    at ContextifyScript.Script.runInContext (vm.js:35:29)
```

But, I've also tested `ember-cli-page-object@master` within our application and can confirm that it's serving in FastBoot without any problem 🎉.

san650:
@bobisjan thanks for the heads up.

Still not clear to me if we need to change our implementation to avoid including the addon on a fastboot build. When you say master is working, are you referring to master with this PR merged?

bobisjan:
The error I've mentioned was fixed in `master` by #346, not by this PR.

I don't know which version of `ember-cli-page-object` @demersus was using, but I can imagine that #346 also fixed his problem.

In FastBoot 1.0+ addon's code can not be actually excluded, but it can be wrapped with `if` statement to prevent from being evaluated, https://github.com/ember-fastboot/ember-cli-fastboot#build-hooks-for-fastboot.

I don't think that it is required to change implementation (because it's working), but it would be useful to have some basic test to prevent from regression using https://github.com/tomdale/ember-cli-addon-tests.

san650:
Thanks @bobisjan for the explanation. I've created the an issue to create a test for fastboot. See #362

PR #351, opened 2017-12-15
feat(focus): Adds focusable and blurrable helpers

pzuraq:
`triggerable` works pretty well for most cases with generic events, and historically it has been the recommended path for focussing and blurring an element. However, unlike most normal triggerable events, focus and blur actually change extra state such as `document.activeElement`, and different contexts may-or-may-not handle this when triggering the event alone. By adding `focusable` and `blurrable` helpers we can ensure that all of the proper steps are taken to trigger these events in the correct way for every context.


addon/-private/dsl.js
```diff
@@ -1,8 +1,10 @@
 import { as } from './properties/as';
+import { blurrable } from './properties/blurrable';
```
san650: I think you forgot to include these files in the PR
* addon/-private/properties/blurrable.js
* addon/-private/properties/focusable.js

pzuraq: That would explain it 😅 updated!

san650: nevermind! I've just found them.

addon/-private/properties/blurrable.js
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * @public
+ *
+ * @param {string} selector - CSS selector of the element which will be blurred
+ * @param {Object} options - Additional options
+ * @param {string} options.scope - Nests provided scope within parent's scope
+ * @param {number} options.at - Reduce the set of matched elements to the one at the specified index
+ * @param {boolean} options.resetScope - Ignore parent scope
+ * @param {string} options.testContainer - Context where to search elements in the DOM
+ * @param {string} options.eventProperties - Event properties that will be passed to trigger function
```
ro0gr: Looks like `eventProperties` are not used in this helper

addon/-private/properties/focusable.js
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * @public
+ *
+ * @param {string} selector - CSS selector of the element which will be focused
+ * @param {Object} options - Additional options
+ * @param {string} options.scope - Nests provided scope within parent's scope
+ * @param {number} options.at - Reduce the set of matched elements to the one at the specified index
+ * @param {boolean} options.resetScope - Ignore parent scope
+ * @param {string} options.testContainer - Context where to search elements in the DOM
+ * @param {string} options.eventProperties - Event properties that will be passed to trigger function
```
ro0gr: Looks like `eventProperties` are not used in this helper

addon/-private/properties/blurrable.js
pzuraq: ah good point, yeah it's not really something you can/would want to pass along. I'll remove it from the docs

ro0gr:
These changes lgtm. 
I think it's nice to have advanced focus/blur on board 👍

addon/-private/properties/focusable.js
```diff
@@ -0,0 +1,74 @@
...
+  return {
+    isDescriptor: true,
+
+    get(key) {
+      return function() {
+        const executionContext = getExecutionContext(this);
+        const options = assign({ pageObjectKey: `${key}()` }, userOptions);
+
+        return executionContext.runAsync((context) => {
+          context.focus(selector, options);
```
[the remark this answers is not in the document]
pzuraq: I think that would be very helpful! It looks like there isn't a single way to determine if an element is focusable, but [according to this Stack Overflow post](https://stackoverflow.com/questions/18261595/how-to-check-if-a-dom-element-is-focusable) it should be possible with a few checks against the various possible things that make an element focusable. We can implement it with these options and expand it later on if we find that there are other things that make an element focusable.

addon/-private/execution_context/helpers.js
```diff
@@ -17,7 +17,7 @@ import $ from '-jquery';
  * @param {string} options.pageObjectKey   Key of method on PageObject which, when invoked, resulted in this call to `fillElement`
  * @return
  *
- * @throws Will throw an error if called on a contenteditable element that has `contenteditable="false"`
+ * @throws Will throw an error if called on a contenteditable element that has `contenteditable='false"`
```
ro0gr: something went wrong here 😃

pzuraq: Oops, good catch. Must've been search and replace

addon/-private/execution_context/helpers.js
```diff
@@ -36,3 +36,37 @@ export function fillElement(selection, content, { selector, pageObjectNode, page
     $selection.val(content);
   }
 }
+
+/**
+ * @private
+ *
+ * Given an element, asserts that element is focusable/blurable
+ *
+ * @param {jQuery} element - jQuery object containing the element(s) to check
```
magistrula: I don't feel super strongly, but I generally prefer calling this `$selection` or `$collection` or something. I tend to think when I see a variable or param called `element` that someone is calling it out as a native DOM element as opposed to a jQuery object, and that's generally the case in this codebase.

tests/unit/-private/properties/focusable-test.js
```diff
@@ -0,0 +1,210 @@
...
+    });
+
+    this.adapter.createTemplate(this, page);
+
+    this.adapter.throws(assert, function() {
+      return page.foo.bar.baz.qux();
+    }, /page\.foo\.bar\.baz\.qux/, 'Element not found');
+  });
+
+  test('raises an error when the element is not focusable', function(assert) {
```
magistrula: I suppose to be really thorough, we would need another block like this with assertions confirming that elements that should be focusable don't throw an error.

addon/-private/execution_context/acceptance.js
```diff
@@ -60,6 +61,30 @@ AcceptanceExecutionContext.prototype = {
     triggerEvent(selector, container, eventName, eventOptions);
   },
 
+  focus(selector, options) {
+    let element = this.findWithAssert(selector, options)
```
magistrula: Suggest the following, for consistency with existing code (see [here](https://github.com/san650/ember-cli-page-object/pull/351/files#diff-49dc265ae7bed20f928377a296d9c3a7R94) and [here](https://github.com/san650/ember-cli-page-object/blob/master/addon/-private/execution_context/native-events-context.js#L73), for example):
```
let element = this.findWithAssert(selector, options)[0];
```
or
```
let $selection = this.findWithAssert(selector, options);
```

addon/-private/execution_context/helpers.js
```diff
@@ -36,3 +36,37 @@ export function fillElement(selection, content, { selector, pageObjectNode, page
...
+  let error;
+
+  let tabIndex = +$element.attr('tabindex');
+  tabIndex = isNaN(tabIndex) ? -1 : tabIndex;
+
+  if ($element.is(':hidden')) {
+    error = 'hidden';
+  } else if ($element.is(':disabled')) {
+    error = 'disabled';
+  } else if (!$element.is(':input, a[href], area[href], iframe') && !(tabIndex > -1)) {
```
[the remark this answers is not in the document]
pzuraq: According to the SO post you linked it looks like `tabindex=-1` is not focusable, but will add the `contenteditable` and `select` types 👍

addon/-private/execution_context/helpers.js
```diff
@@ -36,3 +36,37 @@ export function fillElement(selection, content, { selector, pageObjectNode, page
...
+
+/**
+ * @private
+ *
+ * Given an element, asserts that element is focusable/blurable
+ *
+ * @param {jQuery} element - jQuery object containing the element(s) to check
+ */
+export function assertFocusable(element, { selector, pageObjectNode, pageObjectKey }) {
+  let $element = $(element);
```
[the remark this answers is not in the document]
pzuraq: I was mainly following the convention in `fillElement` that wraps the selection for safety (I assume)  but I'm ok with dropping the extra wrapping, we are currently always passing in a jquery object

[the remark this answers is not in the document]
pzuraq: Ah, actually in `native-events-context` we do just pass the actual element forward. It would be possible to leave it unwrapped I think, but that's why `fillElement` also wraps the element in jquery. I think it'd be best to be consistent here, but I'm open. Thoughts?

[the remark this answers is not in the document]
magistrula: `fillElement` has some technical debt: we updated it because it needed to be able to take a native DOM element instead of a jQuery selection, and we intended to go through and clean up the remaining uses for consistency/simplicity so that we only ever pass an `element` and never pass a `$selection`. cc: @ro0gr

addon/-private/execution_context/helpers.js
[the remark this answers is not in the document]
magistrula: Just to close this loop, @pzuraq and I poked around a bit more and found out that parts of the SO post are incorrect...any tabindex value makes an element focusable: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

addon/-private/execution_context/helpers.js
```diff
@@ -36,3 +36,34 @@ export function fillElement(selection, content, { selector, pageObjectNode, page
...
 }
+
+/**
+ * @private
+ *
+ * Given an element, asserts that element is focusable/blurable
+ *
+ * @param {Element} element - the element to check
+ */
+export function assertFocusable(selection, { selector, pageObjectNode, pageObjectKey }) {
```
magistrula: This first param is `element` now, right?

addon/-private/execution_context/helpers.js
```diff
@@ -36,3 +36,34 @@ export function fillElement(selection, content, { selector, pageObjectNode, page
...
+  let $selection = $(selection);
+
+  let error;
+
+  if ($selection.is(':hidden')) {
+    error = 'hidden';
+  } else if ($selection.is(':disabled')) {
```
magistrula: Let's add an additional block for `[contenteditable="false"]` so that `[contenteditable]` on line 56 excludes elements that have the `contenteditable` attr but which are not actually editable.

tests/unit/-private/properties/blurrable-test.js
```diff
@@ -0,0 +1,249 @@
...
+    });
+
+    this.adapter.createTemplate(this, page);
+
+    this.adapter.throws(assert, function() {
+      return page.foo.bar.baz.qux();
+    }, /page\.foo\.bar\.baz\.qux/, 'Element not found');
+  });
+
+  test('Does not assert on focusable elements', function(assert) {
```
magistrula: Maybe "Does not raise error when blurring focusable elements"? (same suggestion in `focusable` tests).

addon/-private/execution_context/helpers.js
```diff
@@ -36,3 +36,36 @@ export function fillElement(selection, content, { selector, pageObjectNode, page
...
+
+  let error;
+
+  if ($element.is(':hidden')) {
+    error = 'hidden';
+  } else if ($element.is(':disabled')) {
+    error = 'disabled';
+  } else if ($element.is('[contenteditable="false"]')) {
+    error = 'contenteditable="false"';
+  } else if (!$element.is(':input, select, a[href], area[href], iframe, [contenteditable], [tabindex]')) {
```
magistrula: Ah, you were right to leave out `select` the first time around, @pzuraq ... they'll be included in `:input`. I'll remove this when I push those changes to the tests.

PR #357, opened 2017-12-29
Support RFC268 tests

### LINKED ISSUE #355: Support new QUnit modules "acceptance testing"
Using the new QUnit modules API, "acceptance tests" (which aren't exactly a thing anymore) don't register the global test helpers `visit`, `click`, etc. Instead such tests import the helpers explicitly from `ember-test-helpers`. Here's the blueprint:

```javascript
import { module, test } from 'qunit';
import { visit, currentURL } from '@ember/test-helpers';
import { setupApplicationTest } from 'ember-qunit';

module('blah blah blah', function(hooks) {
  setupApplicationTest(hooks);

  test('visiting /some-route', async function(assert) {
    await visit('/some-route');

    assert.equal(currentURL(), '/some-route');
  });
});
```

Currently we can't use a page objects because in acceptance test mode, the helpers are unavailable.

bendemboski:
I think this PR is best read commit by commit rather than as a single changeset.

This implements an execution context for [RFC268](https://github.com/emberjs/rfcs/blob/master/text/0268-acceptance-testing-refactor.md)-style tests, meaning tests that use `setupApplicationTest`/`setupRenderingTest` instead of `moduleForAcceptance`/`moduleForComponent`. It's not required, but this execution context will work much better ergonomics-wise with [async/await](http://rwjblue.com/2017/10/30/async-await-configuration-adventure/) than with promise chaining.

The rfc268 execution context uses the `@ember/test-helpers` helpers (`click`, `visit`, `fillIn`, etc.), which use native events.

The rfc268 execution context supports call chaining with the same async behavior as existing tests, i.e. waiting in between chained async calls, but when used without chaining, reflects the RFC268 async behavior, i.e. nothing is async unless explicitly `await`ed.

Still TODO:
- [ ] Update documentation
- [x] Fix test coverage
- [x] Run codemode for Javascript modules API

Fixes #355


ro0gr:
I've just ran [code-corps-ember](https://github.com/code-corps/code-corps-ember) test suite(as an example of pretty extensive test suite using `ember-cli-page-object`) against this branch.  Tests are green. 

Great job!

tests/helpers/properties/application-adapter.js
```diff
@@ -0,0 +1,43 @@
+import $ from 'jquery';
```
ro0gr: as far as I understand `$` here is exactly the same as `Ember.$`. Probably we should replace all the `'jquery'` imports with `'-jquery'`  in tests as well?...

addon/-private/execution_context/rfc268.js
```diff
@@ -0,0 +1,166 @@
...
+  fillIn,
+  triggerEvent
+} from '@ember/test-helpers';
+import {
+  ELEMENT_NOT_FOUND,
+  throwBetterError
+} from '../better-errors';
+import Ceibo from 'ceibo';
+
+const { $, RSVP: { resolve } } = Ember;
```
ro0gr: We should not use `Ember.$` in the `e-c-p-o` internally. That would not allow to use this contexts in the app with disabled `jquery`.

It can be fixed by importing an isolated jquery instance:
https://github.com/san650/ember-cli-page-object/blob/57183487aa9a8446861f4fbc52c03dbc02eeb1e6/addon/-private/execution_context/native-events-context.js#L1

ro0gr: Also we can use new modules imports.
```
import { resolve } from 'rsvp';
```

bendemboski: Yeah, I was wondering about that -- should I just run the codemod to migrate everything to the new module imports?

ro0gr: Sure, i think it quickest way to do it

tests/acceptance/rfc268-actions-test.js
```diff
@@ -63,6 +67,38 @@ module('Acceptance | actions [rfc268]', function(hooks) {
...
+
+  test('allows testing loading behavior using settled()', async function(assert) {
+    await page.visit();
+    await page.keys.clickOn('1');
+    await page.keys.clickOn('2');
+    await page.keys.sum();
+    await page.keys.clickOn('3');
+    page.keys.asyncEqual();
+
+    await waitUntil(() => page.isLoading);
```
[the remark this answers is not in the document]
bendemboski: Yeah, I thought about that too. But then I realized that probably sometimes you want to wait for an element to exist, and sometimes wait for it to be visible. (`@ember/test-helpers`'s `waitFor` just waits for it to exist). So then I thought maybe an optional callback to verify conditions other than just pure existence...then I thought that doesn't seem to really add any value beyond using `waitUntil()` directly:

```javascript
await waitUntil(() => page.foo);
```

or

```javascript
await waitUntil(() => page.isFooVisible);
```

So it's fully doable, but I couldn't convince myself it would add much value...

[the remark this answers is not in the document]
ro0gr: I think the most common use case is to wait for an element visibility. 
So `waitFor` property could always wait for `isVisible` of the node. If the user wants to customize a wait condition he can always use a new good `waitUntil` helper.

However it's totally unrelated to the PR. I've just realized that it can be a good addition to API.

tests/helpers/properties/application-adapter.js
bendemboski: This is something I'm not familiar with -- what's the difference between importing from `jquery` and `-jquery`?

bendemboski:
I think this is all ready to go except for updating the documentation. I'm not quite sure how to do that. Options that I can think of:

1. Keep all the existing examples the same and add a single section somewhere explaining how the RFC268 test API differs (basically just adding `await`s everywhere)
2. Duplicate all the existing examples to also illustrate the new test API syntax
3. Modify all the existing examples to show the new syntax and then link to a frozen version of the existing documentation site for "legacy" tests

I need some direction here...

tests/helpers/properties/application-adapter.js
ro0gr: We can't rely on `Ember.$` cause it can be empty in apps which don't use jquery. 
`ember-cli-page-object` adds its own jquery version in case of app's jquery absence:

https://github.com/san650/ember-cli-page-object/blob/master/index.js#L29-L31

We call this jquery module '-jquery' in order to prevent overlapping with regular 'jquery'.
https://github.com/san650/ember-cli-page-object/blob/57183487aa9a8446861f4fbc52c03dbc02eeb1e6/vendor/shims/ecpo-jquery.js#L22

In case when `Ember.$` is present we also put it into '-jquery' for convenience:
https://github.com/san650/ember-cli-page-object/blob/57183487aa9a8446861f4fbc52c03dbc02eeb1e6/vendor/shims/project-jquery.js#L14

So I think `import from '-jquery'` is recommended way to use jquery inside this project.

---

I'm not sure if we have to change all the `jquery` imports to `-jquery` in  the tests at least until `Ember` drops `jquery` support. But we definitely have to do it in the [public code](https://github.com/san650/ember-cli-page-object/pull/357/files#diff-6940e0eec69729b817c170c2aa6fccc0R1).

addon/-private/execution_context/rfc268.js
```diff
@@ -0,0 +1,163 @@
+import $ from 'jquery';
```
ro0gr: we should use `-jquery` here cause it can be ran in consumers test suites w/o `Ember.$` available

snewcomer:
@bendemboski This is huuuuge.  Amazing work!  

Coming from a point of naivety (and understanding only a tiny fraction of this PR), is it worthwhile to perf test this at all to see if any big regressions (perhaps related to creating the mirror object)?  On very large acceptance test suites, memory leaks and such are things I have encountered before.

bendemboski:
@snewcomer here's my understanding of the memory story. A page object's `Ceibo` tree is a singleton -- it's created by the `PageObject.create()` call, which happens once per page object when the page object's module is first imported. That's why in your test files you just import your page object, rather than importing a class that you have to instantiate in each test file. So it's kinda immune to memory leaks because it's a singleton and is never destroyed and re-created (it's pre-leaked 😄)! Now we're doubling the amount of memory it consumes by creating the mirror tree, but it's created according to the same pattern and is similarly immune to leaking.

Of course I've made airtight logical arguments about code plenty of times and turned out to be wrong (especially in the areas of perf and memory consumption), so I think doing some perf testing to verify is a fine idea, I just don't personally think the likelihood of a regression is high enough that I feel motivated to make time to do such testing...

snewcomer:
Agreed.  Thanks for the great explanation.

ro0gr:
@bendemboski my 2¢ regarding the docs. 

After some reading of docs I can say there are very few examples of `moduleFor*` usages currently. I can find only `moduleForComponent` in a [integration tests](http://ember-cli-page-object.js.org/docs/v1.12.x/quickstart#integration-tests) section on the quickstart page.

That makes me think we could add just a tiny section with few RFC268 examples. We could also explain that with this feature we don't need to deal with `setupContext(this)/teardownContext()` anymore.

> basically just adding awaits everywhere

I think `await` is not very related here. I believe it can be used with current master even w/o RFC268 support.

bendemboski:
@ro0gr `await` can be used in master without these changes, but with these changes, `await`ing (or promise chaining) every helper call is mandatory. Whether in an application (formerly acceptance) or integration (formerly rendering) test, you need to `await` the helpers because they all [wait a tick](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/dom/click.js#L52) (which is a 0ms timer) before actually executing, so

```javascript
page.click('.foo');
assert.ok(fooWasClicked());
```

will always fail. You need to

```javascript
await page.click('.foo');
assert.ok(fooWasClicked());
```

or if you really want to

```javascript
return page.click('.foo').then(() => {
  assert.ok(fooWasClicked());
});
```

(but I don't think we should bother documenting the promise chaining syntax -- anybody that doesn't want to use `async`/`await` with the new test API is on their own).

ro0gr:
Yes, I totally get the requirement to use `await` for new contexts. I just mean that existing examples are still valid for tests that use old school `moduleFor`. I believe `moduleFor` style remains the default option for v1.x.x of `ember-cli-page-object` so existing examples can be left untouched. We can just focus on writing a section dedicated to RFC268 for now. 

Though I don't feel very strong about that, that's just my understanding.

bendemboski:
Oh yeah, gotcha. Yeah, that makes sense. I agree with treating `moduleFor` tests as the default/common case and RFC268 as the special case for now. Of course at some point (not exactly sure when) `ember-cli` is going to switch the default test blueprints to be RFC268 tests, and at some point more users than not be will running RFC268...but I think you're right that for now it makes the most sense to leave all the documentation the same, but add a section that specifically describes the differences when using the RFC268 API.

addon/-private/execution_context.js
```diff
@@ -1,18 +1,28 @@
...
 /*
  * @private
  */
 export function getExecutionContext(pageObjectNode) {
   let testContext = getContext(pageObjectNode);
-  let context = testContext ? 'integration' : 'acceptance';
+  let context;
+  if (testContext) {
+    context = 'integration';
+  } else if (getTestContext()) {
```
magistrula: Could we add some explanatory comments here? Specifically, how does the return value of `getTestContext()` determine whether we should use the `rfc268` context?

bendemboski:
@san650 @magistrula I just pushed a commit to fix the version/upgrade issues. The solution was to pull `@ember/test-helpers` and `ember-test-helpers` from `package.json`, and then do runtime detection of whether a new enough `@ember/test-helpers` is installed when picking execution contexts. I put together a test project that uses `ember-try` to verify that the difference scenarios that we care about work -- you can check it out [here](https://github.com/bendemboski/ecpo-upgrade-test) if you like.

I also rebased onto master and updated for focus/blur support.

I think maintaining this PR when the test syntax diverges so much from master is going to quickly get very cumbersome, so if we're not going to figure out the documentation question and get this pushed through pretty soon, I'd like to break off the `ember-cli` upgrade and test syntax update (basically the first two commits, but with a little cleanup related to `ember-test-helpers`) into its own PR in such a way that it wouldn't affect users, but would get all the tests on the new syntax, and then get that merged soon.

Does that sound doable?

addon/-private/rfc268-helpers.js
```diff
@@ -0,0 +1,36 @@
...
+
+if (require.has('@ember/test-helpers')) {
+  helpers = require('@ember/test-helpers');
+} else {
+  helpers = {
+    getContext() {
+      return null;
+    }
+  }
+}
```
magistrula: I think it would be helpful for this file to have something like this, to simplify the logic/limit the concerns of `-private/execution_context.js`.
```javascript
export function hasRfc268TestContext() {
  // `helpers.getContext()` returns:
  //  - `null`, if @ember/test-helpers is not available
  //  - _____?, if @ember/test-helpers is available but none of the `ember-qunit` 
  //    setupTest() methods has been called (e.g., `setupRenderingTest()`)
  //  - qunit test context, if @ember/test-helpers is available and one of the
  //    `ember-qunit` setupTest() methods has been called.
  let hasValidTestContext = !!helpers.getContext();

  if (!hasValidTestContext) {
    return false;
  }

  let hasExpectedTestHelpers = !!helpers.visit;  
  if (!hasExpectedTestHelpers) {
      throw new Error([
        'You are trying to use ember-cli-page-object with RFC232/RFC268 support',
        '(setupRenderingContext()/setupApplicationContext()) which requires at',
        'least ember-qunit@3.2.0 or ember-mocha@0.13.0-beta.3.'
      ]);
    }

  return true;
}
```

Then, in `-private/execution_context.js`:
```javascript
export function getExecutionContext(pageObjectNode) {
  let pageObjectContext = getContext(pageObjectNode);
  let context;

  if (pageObjectContext) {
    context = 'integration';
  } else if (hasRfc268TestContext()) {
    context = 'rfc268';
  } else {
    context = 'acceptance';
  }

  return new executioncontexts[context](pageObjectNode, pageObjectContext);
}
```

addon/-private/execution_context.js
```diff
@@ -1,18 +1,43 @@
...
+  // `setupApplicationTest()`). If neither of these are the case, the only
+  // supported scenario is an acceptance test.
   let testContext = getContext(pageObjectNode);
-  let context = testContext ? 'integration' : 'acceptance';
+  let context;
+  if (testContext) {
+    context = 'integration';
+  } else if (getRfc268Context()) {
+    if (!visit) {
+      throw new Error([
```
magistrula: Under what circumstances could we end up here? When would `getRfc268Context()` return `true` while `visit` is undefined?

bendemboski: There were a few minor versions of `@ember/test-helpers` that had `setupRenderingTest()` implemented, but did not have the DOM helpers. In one of my projects, I upgraded `ember-qunit` to start implementing RFC232 rendering tests using `ember-cli-page-object`, but still using `page.setContext()` so it was using the integration execution context.

Since it's in my `yarn.lock`, if I were to update to this version of `ember-cli-page-object` and delete the `page.setContext()`s to try to use the RFC268 execution context, I'd get cryptic `helper is not a function` (or something) errors, so this is meant to clarify what's going on.

There was also a brief window when all the helpers but `visit()` were implemented, but nobody using `ember-cli-page-object` could have actually been using them since there was no execution context that called them. So to keep things clean and simple (ha!), we check for `visit()` which was the last one to be added and guarantees that you can run both RFC232 tests *and* RFC268 tests.

addon/-private/rfc268-helpers.js
bendemboski: Yeah, I guess I could do that, but there's not a really clean separation of concerns. The logic in `hasRfc268TestContext()` relies on the fact that we've already ruled out the integration context in order to be sure it's making the right compatibility/upgrade decisions for the user. We have to make `page.setContext()` force integration mode or we break compatibility with people that have started using `ember-cli-page-object` with RFC232 tests (which works fine on master).

So the logic is really a cascade of decisions:

1. Integration test or RFC232 test that is calling `page.setContext()`? => integration
1. RFC232 or RFC268 test with `visit` present? => rfc268
1. else => acceptance

and it seems to me that the importance of the ordering is harder to express when the decision logic isn't all in the same file.

I don't feel strongly about this, so I'm happy to move the code like you suggest, but my personal judgement is that the loss of locality takes away more from the readability than the (imperfect) separation of concerns adds. But say the word and I'll happily concede the point 😄

addon/-private/helpers.js
```diff
@@ -1,8 +1,11 @@
-import Ember from 'ember';
+//import { assign, merge } from '@ember/polyfills';
```
ro0gr: ✂️

addon/-private/helpers.js
```diff
@@ -310,4 +313,4 @@ export function getProperty(object, pathToProp) {
   return typeof value === 'function' ? value.bind(propOwner) : value;
 }
 
-export const assign = Ember.assign || Ember.merge;
+//export const assign = assign || merge;
```
ro0gr: ✂️

addon/-private/rfc268-helpers.js
magistrula: Yeah, I say let's go ahead and make these changes. I think these explanatory comments and error handling belong in this file rather than `-private/execution_context.js`.

bendemboski: Okay, pushed those changes (sorry it took a bit -- was on vacation and not connected for the long weekend)

addon/-private/execution_context.js
```diff
@@ -1,18 +1,38 @@
...
+  // Our `getContext(pageObjectNode)` will return a context only if the test
+  // called `page.setContext(this)`, which is only supposed to happen in
+  // integration tests (i.e. pre-RFC232/RFC268). However, the integration
+  // context does work with RFC232 (`setupRenderingContext()`) tests, and before
+  // the RFC268 execution context was implemented, some users may have migrated
+  // their tests to RFC232 tests, leaving the `page.setContext(this)` in place.
+  // So, in order to not break those tests, we need to check for that case
+  // first, and only if that hasn't happened, check to see if we're in an
+  // RFC232/RFC268 test, and if not, fall back on assuming a pre-RFC268
+  // acceptance test, which is the only remaining supported scenario.
```
magistrula: This is really clear; awesome.

addon/-private/is-rfc268-test.js
```diff
@@ -0,0 +1,37 @@
...
+  if (!hasValidTestContext) {
+    return false;
+  }
+
+  // There are a few versions of `@ember/test-helpers` that have support for
+  // `ember-qunit`'s `setupRenderingTest()` method, but do not have the DOM
+  // helpers (`click`, `fillIn`, etc.) that the RFC268 execution context uses.
+  // `visit` was the last helper to be added to `@ember/test-helpers`, so we
+  // check for it, and if we can't find it, we can't use the RFC268 execution
+  // context, so we throw an exception.
```
magistrula: Again, super helpful in explaining hard-to-discover subtleties/background information.

addon/-private/helpers.js
```diff
@@ -151,28 +153,28 @@ export function normalizeText(text) {
...
 
-  return Ember.A(arr).map(function(element) {
+  return A(arr).map(function(element) {
     return cb($(element));
   });
 }
 
 /**
- * @private
+ * @public
```
ro0gr: I think technically it's still a private function cause it's not re-exported to the consumer

tests/helpers/properties/rfc268-adapter.js
```diff
@@ -0,0 +1,37 @@
...
+import { settled } from '@ember/test-helpers';
+
+export default function Rfc268Adapter() {}
+
+Rfc268Adapter.prototype = {
+  $(selector, isAlternative) {
+    return $(selector, isAlternative ? '#alternate-ember-testing' : '#ember-testing');
+  },
+
+  async throws(assert, block, expected, message) {
```
ro0gr: not sure if it's the same, but can we use [`assert.rejects`](https://api.qunitjs.com/assert/rejects) here? At least currently we have the latest `qunit` installed which provides us with this api.

tests/helpers/properties/rfc268-adapter.js
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    assert.throws(() => {
+      if (error) {
+        throw error;
+      }
+    }, expected, message);
+  },
+
+  async wait() {
```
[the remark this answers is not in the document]
bendemboski: Oh yeah, that was left over from a previous iteration -- good catch, thanks!

tests/unit/-private/properties/blurrable-test.js
```diff
@@ -171,56 +153,62 @@ moduleForProperty('blurrable', function(test) {
...
-        bar: {
-          input: blurrable('input'),
-          select: blurrable('select'),
-          a: blurrable('a'),
-          area: blurrable('area'),
-          iframe: blurrable('iframe'),
-          button: blurrable('button'),
-          contentEditable: blurrable('[contenteditable]'),
-          tabindex: blurrable('[tabindex]'),
+  if (adapter === 'integration' || adapter === 'acceptance') {
```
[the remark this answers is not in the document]
bendemboski: No, the underlying test helpers don't raise an error if you try to blur and un-focusable element, so this test doesn't just doesn't make sense to run for the new execution context.

tests/unit/-private/properties/blurrable-test.js
```diff
@@ -171,56 +153,62 @@ moduleForProperty('blurrable', function(test) {
...
+      page.foo.bar.button()
+      page.foo.bar.contentEditable();
+      page.foo.bar.tabindex();
+    });
+  }
 
-  test('raises an error when the element is not focusable', function(assert) {
-    assert.expect(4);
+  test('raises an error when the element is not focusable', async function(assert) {
+    if (adapter === 'acceptance' || adapter === 'integration') {
```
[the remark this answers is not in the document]
bendemboski: Personally, I wouldn't use `assert.expect()` in this case since the only way it would fail is if a promise rejected, which would fail the test anyway. But I don't feel comfortable applying that preference to code written so recently, unless @pzuraq agrees that removing it makes sense.

addon/-private/helpers.js
bendemboski: Oh yeah, I thought some of the other `@public` helpers weren't exported, but I see I was mistaken. I'll switch this one back.

bendemboski: A couple of the other helpers that are exported from the file, but not to the consumer, are marked as `@public` like `findClosestValue` and `objectHasProperty` so it seems that the `@public` and `@private` are just referring to whether it's exported from the file or not, not whether it's exported to consumers of the addon.

tests/helpers/properties/rfc268-adapter.js
bendemboski: I don't think it's quite the same...mumble mumble test adapter something something...I can't exactly recall but I think we have to do it just so to keep the test setup working correctly.

tests/unit/-private/properties/blurrable-test.js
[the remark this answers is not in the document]
pzuraq: I did it mostly out of habit, with the older style tests it was very easy to have flakiness because assertions would get dropped in race conditions. I don't think there's a reason to keep it at this point.

bendemboski:
Okay, this is all rebased and resolved against the latest master, and ready to merge pending further reviews/feedback.

bgentry:
Just wanted to drop a note that I used this to upgrade my app to Ember v3.0 and the only changes I had to make were to put `await` in front of each chained call to my page object. Thanks so much for doing this! 👏

alexlafroscia:
I'm trying to test this out in an app that I'm upgrading, but `page.visit` isn't working. It correctly picks the new execution context that was added in this PR, but it tries to call `owner.visit` which is not defined.

Is the intention that we're defining that ourselves, or is some dependency supposed to define that for us?

wayne-o:
Would be good if this could get released since anyone using it with a new ember project will run into problems - is there a timeline for this?

alexlafroscia:
I'm not sure where the bug lies, but trying out this PR over the last day, I've had to add some

```javascript
await wait();
```

after using `clickable` where it was not required before.

san650:
@alexlafroscia was the action defined in a collection? Could it be related to #377?

I don't know if the fix for #377 is included in this branch.

If that's not the case, could you provide an example? So we can triage the error.

bendemboski:
@alexlafroscia are you `await`ing the invocation of the clickable? Can you post a code snippet?

alexlafroscia:
It could definitely be related to that issue @san650, it is defined within a collection.

I'm also seeing an issue with `click`-ing an item within a nested collection (collection item in turn has a collection, where I'm trying to `click` one of those items)

At this line

https://github.com/san650/ember-cli-page-object/pull/357/files#diff-6940e0eec69729b817c170c2aa6fccc0R76

I have a path like `["wizard", "date_range_groups[0]", "categories[0]"]`, which expands out to

````
node['wizard']['date_range_groups[0]']['categories[0]']
```

when what actually works would be something likle

```
node['wizard']['date_range_groups'][0]['categories'][0]
```

bendemboski:
@alexlafroscia nice catch -- I just pushed a commit resolving the issues with `await`ing and chaining calls to actions on collection items.

alexlafroscia:
Awesome! Thanks for that @bendemboski, I pulled the latest from your branch and it's working again 👍

addon/-private/is-rfc268-test.js
```diff
@@ -0,0 +1,37 @@
...
+  let hasExpectedTestHelpers = Boolean(visit);
+  if (!hasExpectedTestHelpers) {
+    throw new Error([
+      'You are trying to use ember-cli-page-object with RFC232/RFC268 support',
+      '(setupRenderingContext()/setupApplicationContext()) which requires at',
+      'least ember-qunit@3.2.0 or ember-mocha@0.13.0-beta.3.'
```
simonihmig: FWIW, that version of `ember-mocha` does *not*  support RFC268-based tests. It might ship with an appropriate version of `@ember/test-helpers`, but it does not yet provide the `setupRenderingTest()/setupApplicationTest()` functions to setup such tests, as `ember-qunit` does...

I know that for sure, because my [PR](https://github.com/emberjs/ember-mocha/pull/190) for adding those just got merged today 😉

bendemboski: Heh, okay, should I optimistically change the version to `0.13.2` then?

addon/-private/compatibility.js
```diff
@@ -1,12 +1,67 @@
...
+  helpers = {
+    getContext() {
+      return null;
     }
-  }
+  };
+}
+
+if (require.has('ember-test-helpers/wait')) {
+  helpers.wait = require('ember-test-helpers/wait').default;
```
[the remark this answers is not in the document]
bendemboski: @simonihmig this is to support legacy apps. `ember-test-helpers`'s `wait` hasn't always existed. `ember-cli-page-object` previously had `ember-test-helpers` as a runtime dependency to be sure `wait` was present. Now it doesn't, so apps with an old enough `ember-qunit`/`ember-test-helpers` won't have the `wait` helper present and will run into that exception you hit instead of the nice descriptive error message that you didn't get :(

I'm confused about exactly what you're seeing. Are you running `ember-cli-page-object`'s unit tests? Or your own tests using this branch? What version of `@ember/test-helpers`/`ember-test-helpers` do you have installed? `loader.js` isn't that complicated -- if you want to peel back an additional layer you could step into the require call and maybe see why it's giving you back an empty object...

addon/-private/is-rfc268-test.js
simonihmig: Hm, not sure what version that would be... @Turbo87 any thoughts?

addon/-private/compatibility.js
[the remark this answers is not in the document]
simonihmig: I am running my apps tests, using the master branch of `ember-mocha`. Should be the latest version of `@ember/test-helpers`, but will double check. 

> if you want to peel back an additional layer you could step into the require call and maybe see why it's giving you back an empty object...

Ok, will do that, but not before tomorrow...

addon/-private/is-rfc268-test.js
Turbo87: it's actually worse than that. even with `ember-qunit@3.2.0` and `visit` available you can't be sure that you can actually use it, because you might be dealing with a testsuite that still uses the old testing APIs and those don't work with the new `visit()` function

ro0gr: @Turbo87 this check occures [only](https://github.com/san650/ember-cli-page-object/pull/357/files/464af458da5d7f45582f21c02732d9c0cf334987#diff-af2391cd4be4b040e0c0eddf7e2d444dR16) when we use the new testing API. I think the question here is which minimal versions of `ember-qunit` `ember-mocha` do provide a working `visit()` helper?

bendemboski: `visit` was added in `@ember/test-helpers@0.7.9`, which was first required by `ember-qunit@3.2.0`, which is why the error references that version of `ember-qunit`.

@ro0gr is correct -- the question is when `ember-mocha` will support it. As far as I understand, @simonihmig's PR to add full support was just merged, so will that be released as `0.13.2`?

addon/-private/execution_context/rfc268.js
```diff
@@ -0,0 +1,194 @@
...
+        // Normally an item's key is just its property name, but collection
+        // items' keys also include their index. Collection item keys look like
+        // `foo[2]` and legacy collection item keys look like `foo(2)`.
+        let match;
+        if ((match = /\[(\d+)\]$/.exec(key))) {
+          // This is a collection item
+          let [ indexStr, index ] = match;
+          let name = key.slice(0, -indexStr.length);
+          node = node[name].objectAt(parseInt(index, 10));
+        } else if ((match = /\((\d+)\)$/.exec(key))) {
```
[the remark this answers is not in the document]
bendemboski: Won't that make migrating pretty painful? I still have a bunch of code using legacy collections that I'm gonna migrate sometime, but haven't had time to yet since they were just released recently. The new test API has be released for quite some time and especially with Ember 3.0, many users are already using it. Do we really want to block users from using the new test API until they migrate all of their collections?

addon/-private/compatibility.js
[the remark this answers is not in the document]
bendemboski: Okay. I haven't seen anything like that, and the `ember-cli-page-object` tests run all the relevant `ember-try` scenarios, so I'm not sure how to investigate with more info or a sample project that reproduces it. So let me know what you find!

addon/-private/execution_context/rfc268.js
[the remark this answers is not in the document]
ro0gr: Do you propose to release this as a part of v1 or to preserve legacy collections support in v2?
 
I think we can't do the former cause we change the way properties are defined which is a part of public API I believe. On the other hand preserving legacy collections support in v2 seems a bit strange(but acceptable at least to me). 
mhm.. Maybe it's a time for collections codemod...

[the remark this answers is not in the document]
san650: > Do you propose to release this as a part of v1 or to preserve legacy collections support in v2?

Migrating to v2 of page object is going to be a lot of work for people like @bendemboski said, I think we should strive for having this PR merged in for v1

> I think we can't do the former cause we change the way properties are defined which is a part of public API I believe

Can you expand a bit on this? I'm not sure to understand what would break.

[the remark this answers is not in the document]
ro0gr: I'm reffering to async props like this https://github.com/san650/ember-cli-page-object/pull/357/files#diff-52672839c703d1e4a3286b560cd21215R78. We should now return an async operations. I suspect this change might break someones custom props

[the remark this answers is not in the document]
ro0gr: Actually I'm not sure how wide-spread is custom props in the wild. 
Also seems like we haven't [documented](http://ember-cli-page-object.js.org/docs/v1.13.x/extend) declaring of async custom properties yet so it might be considered as a private API.

If that's ok I totally support including the new test API to the v1

bgentry:
@bendemboski I was encountering some issues with chaining on collections and with `await click()` not actually waiting, but then I found out you'd already fixed them :pray: All working great after the latest commits 🎉

addon/-private/compatibility.js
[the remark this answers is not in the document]
simonihmig: @bendemboski so here is what I got...

First of all I have `@ember/test-helpers` 0.7.18 installed...

When I step into the require call, I can see that the module is in a `pending` state, and the exports are empty:

image

As stated before, when I explicitly import `ember-test-helpers/wait` in that file, it works, and as you can see the module is in the `finalized` state and the exports are all there:

image

This is happening only in a few tests, so I *suppose*  this is depending on the order the modules are required: when `wait` has been required before by some other module, it is already in the `finalized` state, and the require gets the correct exports. But when the `require` call in question here is the first to ever need the module, and it has not declared `ember-test-helpers/wait` as a dependency (by importing it), it will see a `pending` module and the exports are still empty.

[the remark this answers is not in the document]
bendemboski: I'll try to get a repro of this when I have the chance (probably later today or tomorrow), but if you have time, could you try changing this line to

```javascript
helpers.wait = (...args) => require('ember-test-helpers/wait').default(...args);
```

and see if it works for you? I'm hoping that moving the require to "runtime" rather than "module load time" (if you know what I mean) will fix it.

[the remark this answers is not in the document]
simonihmig: Just tried it quickly, and yes, it indeed fixes the problem! 😀

I probably miss something important, but what would be the problem if `e-c-p-o` would simply add the latest `@ember/test-helpers` as a dependency? (so you could simply import it and not require all of this)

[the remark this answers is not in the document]
bendemboski: It used to. @magistrula observed that (maybe due to an `ember-cli` bug or maybe an intended feature?) if the app also specified a version of `@ember/test-helpers`, the addon's version of `@ember/test-helpers` was overriding the version the app specified (see #361 and #366). This seems like risky behavior to accept in order to patch up a very uncommon case (an app having a version of `ember-test-helpers` in the pretty old and narrow window when it didn't have `wait`).

Also, when I asked in slack, I think it was @Turbo87 and/or @rwjblue that told me that apps and addons can and should just depend on `ember-qunit` or `ember-mocha` to supply `@ember/test-helpers`, and not list it as a dependency themselves.

Anyway, thanks a bunch for digging in -- I'll make that change to defer the require execution until after modules are loaded.

ro0gr:
Thanks to @bendemboski and everyone who took a part here!

san650:
👏 👏 👏 👏 👏 👏 👏 👏  Amazing work everyone!!!

Thanks @bendemboski for tackling this really difficult task! ❤️

addon/-private/helpers.js
```diff
@@ -1,9 +1,11 @@
-import Ember from 'ember';
+export { assign } from '@ember/polyfills';
```
vitch: I think this introduced a dependency on a particular version of ember.

In our apps (currently using ember `2.16.2`) I got an error when I upgraded to `ember-cli-page-object` `1.15.0-beta.2` (so that I could use the goodness from this PR):

> Could not find module `@ember/polyfills` imported from `ember-cli-page-object/-private/helpers`

I can't remember is `assign` from Ember does anything special but I tried replacing this line with:

```js
export const assign = Object.assign;
```

locally and it *seems* to be working...

Turbo87: can you reproduce that in a fresh `ember new` app? seems strange to me because the `ember-cli-babel` dependency from the addon should take care of transpiling that correctly 🤔

vitch: Thanks for the hint :)

I tried to reproduce it in a fresh `ember new` app based on the same version of `ember-cli` and... I can't! See https://github.com/vitch/test-page-object/tree/test-page-object

I'm going to dig further to try and find out what we are doing which is causing problems and I'll let you know if I manage to reproduce it...

PR #360, opened 2018-01-06
Focusable/Blurrable tweaks

magistrula:
### Purpose
- Remove unneeded check for `select` elements in `assertFocusable`. `select` elements are handled by the `:input` check.
- Add assertions for the focusability  of `contenteditable="false"` divs.

### DIFF
--- a/tests/unit/-private/properties/blurrable-test.js
+++ b/tests/unit/-private/properties/blurrable-test.js
@@ -204,7 +204,7 @@ moduleForProperty('blurrable', function(test) {
       <iframe></iframe>
       <select></select>
       <button></button>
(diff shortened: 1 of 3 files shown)


tests/unit/-private/properties/blurrable-test.js
```diff
@@ -204,7 +204,7 @@ moduleForProperty('blurrable', function(test) {
       <iframe></iframe>
       <select></select>
       <button></button>
-      <div contenteditable=true></div>
+      <div contenteditable></div>
```
magistrula: Removed the `true` here, because the code only cares about `contenteditable="false"`, so the `true` is not needed.

PR #363, opened 2018-01-07
Update documentation command

magistrula:
### Purpose
We updated the `documentation` module to `v5.3.3`, and now we need to update the command that builds the docs. 

### Open Questions
- [ ] What did the `--github` argument do? It does not seem to be supported in `5.3.3`.

**Sanity Check**:
After making these updates, I was able to successfully generate the docs that we will publish for `v1.13.0`.


san650:
I'm almost sure that `--github` was used to enable github flavored markdown. It seems that this is the default behavior now.

Issue #361, opened 2018-01-06
ember-cli-page-object clobbers host app version of ember-test-helpers


magistrula:
Observed in an app that uses `ember-test-helpers@0.6.3`, when testing the changes for #357:

`import wait from 'ember-test-helpers/test'` pulls in `wait` from `v0.7.13` (`ember-cli-page-object` version) instead of from `v0.6.3` (host-app version).

san650:
After discussing with @magistrula we came to the conclusion that instead of adding `ember-test-helpers` as a dependency of ember-cli-page-object, we need to move it as a dev-dependency (because of this addon's tests) and install the dependency on the host app using the installation blueprint.

https://ember-cli.com/extending/#working-with-dependencies

This way we'll update host's `ember-test-helpers` dependency if needed.

cc/ @bendemboski

bendemboski:
@san650 @magistrula 

I chatted with some of the Ember core team folks, and they recommend something a little different. The intent is that apps won't specify their own version of `ember-test-helpers` (now called `@ember/test-helpers`), but will just use the version pulled in by `ember-qunit` or `ember-mocha` (which is what your project was doing @magistrula -- apparently correctly, sorry for the bum steer).

So `ember-cli-page-object` really shouldn't have it as a dependency _or_ install it in the host app using the blueprint. If anything, we should be checking for a new enough `ember-qunit` or `ember-mocha` version at runtime, although I don't think that's really necessary unless/until we deprecate pre-RFC268 tests.

So to sum it up, it sounds like we should just simply remove it -- anybody running tests will already have the version that they need installed via `ember-qunit` or `ember-mocha`.

Make sense?

PR #366, opened 2018-01-22
Remove ember-test-helpers from dependencies

magistrula:
### Purpose
- Remove `ember-test-helpers` from `dependencies` and rely on the host app to supply the test helpers, either via `ember-mocha` or `ember-qunit`. This will prevent `ember-cli-page-object` from clobbering the host app's version of the helpers. 
- Addresses https://github.com/san650/ember-cli-page-object/issues/361

### Approach
- Remove `ember-test-helpers` from the list of `dependencies`
- Instead of importing `ember-test-helpers/wait`, create a new `wait` helper that throws an error if `ember-test-helpers` or `@ember-test-helpers` is not available.

### DIFF
--- a/addon/-private/compatibility.js
+++ b/addon/-private/compatibility.js
@@ -1,0 +1,12 @@
+export const wait = (function() {
+  const hasLegacyEmberTestHelpers = window.require.has('ember-test-helpers');
+  const hasLatestEmberTestHelpers = window.require.has('@ember/test-helpers');
+
+  if (!hasLegacyEmberTestHelpers && !hasLatestEmberTestHelpers) {
+    return function() {
+      throw new Error('ember-test-helpers or @ember/test-helpers must be installed');
+    }
+  }
+
+  return window.require('ember-test-helpers/wait')['default'];
+})();
--- a/addon/-private/dsl.js
+++ b/addon/-private/dsl.js
@@ -10,7 +10,7 @@ import { isPresent } from './properties/is-present';
 import { isVisible } from './properties/is-visible';
 import { text } from './properties/text';
 import { value } from './properties/value';
-import wait from 'ember-test-helpers/wait';
+import { wait } from './compatibility';
 
 const thenDescriptor = {
   isDescriptor: true,
--- a/addon/-private/execution_context/acceptance-native-events.js
+++ b/addon/-private/execution_context/acceptance-native-events.js
@@ -1,5 +1,5 @@
 import ExecutionContext from './native-events-context';
-import wait from 'ember-test-helpers/wait';
+import { wait } from '../compatibility';
 
 import {
   visit
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -92,6 +92,38 @@ module.exports = {
       }
     },
     {
+      name: 'with-ember-test-helpers',
+      bower: {
+        dependencies: {
+          'ember': 'components/ember#release'
+        },
+        resolutions: {
+          'ember': 'release'
+        }
+      },
+      npm: {
+        devDependencies: {
+          'ember-cli-qunit': '4.0.0'
+        }
+      }
+    },
+    {
+      name: 'with-@ember/test-helpers',
+      bower: {
+        dependencies: {
+          'ember': 'components/ember#release'
+        },
+        resolutions: {
+          'ember': 'release'
+        }
+      },
+      npm: {
+        devDependencies: {
+          'ember-cli-qunit': '4.3.0'
+        }
+      }
+    },
+    {
       name: 'node-tests',
       command: 'npm run nodetest',
       bower: {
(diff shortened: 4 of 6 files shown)


addon/-private/helpers.js
```diff
@@ -311,3 +315,12 @@ export function getProperty(object, pathToProp) {
 }
 
 export const assign = Ember.assign || Ember.merge;
+
+export function wait() {
+  const { settled, wait } = emberTestHelpers;
```
ro0gr: I think it will fail if `require.has('ember-test-helpers') === false`

magistrula: Ha, good point. I'll go back to my original version:
```
if (emberTestHelpers) {
  ...
}
```

magistrula: Still trying to figure out how to write tests for this.

ro0gr:
@magistrula do you have any ideas how to test this? 
Just an idea: I think we can setup a separate ember-try scenario in order to install some old `ember-cli-qunit` version which will bring `ember-test-helpers` without `settled` provided? or we can even install a specific `ember-test-helpers` version in this scenario...

magistrula:
@ro0gr, that's a good idea, thanks. I'll try it out and remove the "WIP" label from this PR once I figure out how to test.

addon/-private/helpers.js
```diff
@@ -1,9 +1,13 @@
 import Ember from 'ember';
 import Ceibo from 'ceibo';
+import require from 'require';
+import $ from '-jquery';
 
 const { assert, get, isPresent } = Ember;
 
-import $ from '-jquery';
+const emberTestHelpers = require.has('ember-test-helpers') ?
```
bendemboski: I'm pretty sure this won't work for `ember-test-helpers` `0.6.x` and earlier. `ember-test-helpers` has never exported `wait` from the [module index](https://github.com/emberjs/ember-test-helpers/blob/v0.6.3/lib/ember-test-helpers.js), it's always only been available as the default export of `ember-test-helpers/wait`. I think this is the matrix:

### <= 0.6.x
* `require('ember-test-helpers').wait` doesn't exist
* `require('ember-test-helpers/wait')` exists
* `require('ember-test-helpers/settled')` does not exist

### 0.7.x
* `require('ember-test-helpers').wait` does not exists
* `require('ember-test-helpers/wait')` exists
* `require('ember-test-helpers/settled')` exists, but I think kinda by accident since `settled` is part of the new `@ember/test-helpers` API

I think you should probably just do

```javascript
const wait = require.has('ember-test-helpers/wait') ? require('ember-test-helpers/wait') : null;
```

and then check for `wait` being truthy.

magistrula: Ah, good point...thx!

addon/-private/compatibility.js
```diff
@@ -0,0 +1,15 @@
...
+
+function _hasLatestEmberTestHelpers() {
+  return require.has('@ember/test-helpers');
+}
+
+if (!_hasLegacyEmberTestHelpers() && !_hasLatestEmberTestHelpers()) {
+  throw new Error('ember-test-helpers or @ember/test-helpers must be installed');
+}
+
+export { default as wait } from 'ember-test-helpers/wait';
```
bendemboski: Maybe this is too nitpicky and not a case we care about, but this will throw in a technically valid scenario. If the user is only using `ember-cli-page-object` for acceptance tests, then we actually don't need `ember-test-helpers` or `@ember/test-helpers` as the acceptance context uses `window.wait()` which comes from Ember itself.

If that is a case we want to continue supporting, then my suggestion would be to export a function that runs the checks before calling through to `wait`, so we only check and assert when the `wait` functionality is being used, rather than when the modules are loading. Or run the checks at module scope and store the result in a variable, and then check and throw inside the exported function if we're worried about the perf implications of running `require.has()` twice on every `wait()` call.

addon/-private/compatibility.js
```diff
@@ -0,0 +1,12 @@
+export const wait = (function() {
+  const hasLegacyEmberTestHelpers = window.require.has('ember-test-helpers');
+  const hasLatestEmberTestHelpers = window.require.has('@ember/test-helpers');
```
magistrula: The only way I could come up with to test the error case was to use the `window.require` provided by Ember. Is there any reason not to do this?

bendemboski: You mean because of the stubbing, right? I think it would work to do `import require from 'require'` in this code, and then:

```javascript
import require from 'require';

test('blah blah', function(assert) {
  require.has = () => false;
  // run test
});
```

Did you try that?

magistrula: Yeah, that was my first approach. No dice.

bendemboski: What went wrong? That's confusing to me because when I run a test that does `import require from 'require'`, `window.require === require` evaluates to `true`...so I'm not sure how the two are different? Unless you were doing `const { has } = require` or something in `compatibility.js`?

magistrula: Hmmm, I'm not seeing `window.require === require` evaluating to true...I believe Ember provides its own `window.require`, which is different from the imported `require`.

Issue #359, opened 2018-01-06
Planning v2 release


san650:
(This are some ideas and we're open to suggestions)

## Main goals

* Make it easier to maintain
  * Support fewer versions of ember/ember-cli (just 3.0 and above?)
* Remove jQuery dependency
* Support for RFC268 tests (see #357)
* Clean up API
  * New collection API (see #352, #347 and #345)
  * Cleanup API options, helpers, requires, etc.
* Site redesign with ember-cli-addon-docs (#353)
* Migration path from v1.14
* Dedicated deprecations page in the guide (#465) 
* Adapters (#479) 
* Embroider compatibitlity (https://github.com/san650/ember-cli-page-object/issues/539)

## Deprecations

* [x] v1.4 will include the new collection API with deprecation warnings
* [x] [multiple selectors (group)](https://github.com/san650/ember-cli-page-object/issues/335#issuecomment-344248732)
  * [x] v1: deprecate 
  * [x] v2: replace deprecated code with error
* [x] Drop `page.render()`
* [X] imports from test-support https://github.com/san650/ember-cli-page-object/pull/368
* [x] `is` https://github.com/san650/ember-cli-page-object/issues/442
* [x] deprecate findElement/findElementWithAssert #463 
* [x] deprecate `url` in `create` https://github.com/san650/ember-cli-page-object/issues/464
* [x] deprecate `setContext` (#448)
* [x] deprecate `multiple`
* [x] `string-properties-on-definition` deprecation 
  * [x] v1: deprecate 
  * [x] v2: replace deprecation with an error
* [x] `create-url-argument` deprecation 
  * [x] v1: deprecate 
  * [x] v2: remove deprecated code
* [x] update "/docs/guides". Make sure migration page is up to date

ro0gr:
> Remove jQuery dependency

I think in order to achieve that in v2, we should plan deprecations for APIs which are directly bound to jquery, like `is`, `findElement`, and probably others

yratanov:
can I help with this?

ro0gr:
@yratanov Thanks for the interest!

I believe there are few more things needed. I'll work to create missing tickets and update this one today.

SergeAstapov:
@ro0gr @san650 out of curiosity, did you change plans regarding V2 or do you consider it "close to be done"?

ro0gr:
[...]

PR #367, opened 2018-01-24
deprecate page.render()

ro0gr:
part of [1.x deprecations](https://github.com/san650/ember-cli-page-object/issues/359)

TODO:
 - [x] need to find and fix all the `page.render()` usages internally. 
I haven't found any usages.

### DIFF
--- a/addon/-private/context.js
+++ b/addon/-private/context.js
@@ -1,3 +1,5 @@
+import { deprecate } from '@ember/application/deprecations';
+
 /**
  * @public
  *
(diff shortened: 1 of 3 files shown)


addon/-private/context.js
```diff
@@ -17,6 +19,11 @@
  * @return {PageObject} - the page object
  */
 export function render(template) {
+  deprecate('PageObject.render() is deprecated. Please use "htmlbars-inline-precompile" instead.', false, {
```
magistrula: Can we provide a link here to some documentation of what it means to use `htmlbars-inline-precompile`?

ro0gr: Yeah, seems like the message wants to be improved.

I've been thinking about providing an [`url`](https://github.com/emberjs/ember.js/blob/bdc2c3e3a15d13d8bb1951d44b1a28ee1c6a36cb/packages/ember-debug/lib/deprecate.js#L155) deprecation option pointing to some deprecation specific page or section. There we can explain a migration path. 

The question is what this page/section should be? I can see 2 options:
 - deprecations page on the site
 - separate github issue per each deprecation

I personally like the second option cause it doesn't require to restructure the site. Also separate issues give us an ability to receive feedback from the users in case of related questions. What do you think?

magistrula: I think an issue per deprecation showing before/after examples would be perfect.

PR #368, opened 2018-01-24
Deprecate imports from test-support

### LINKED ISSUE #365: Deprecate imports from `./page-object`
In the [docs](http://ember-cli-page-object.js.org/docs/v1.13.x/migrating) we still can see an old style of imports from the `test-support` directory:
```js
import { findElement } from './page-object';
```

I think it's a good time to notify users about the recommended way to import 
```js
import { findElement } from 'ember-cli-page-object';
```
and to get rid of old style in the v2.0.

_This issue is complementary to #359_

ro0gr:
closes #365

### DIFF
--- a/test-support/page-object.js
+++ b/test-support/page-object.js
@@ -1,3 +1,5 @@
+import { deprecate } from '@ember/application/deprecations';
+
 import {
   alias,
   attribute,
@@ -75,3 +77,8 @@ export default {
 };
 
 export { buildSelector, findElementWithAssert, findElement, getContext, fullScope } from 'ember-cli-page-object';
+
+deprecate(`Importing from "test-support" is now deprecated. Just import directly from the "ember-cli-page-object" module instead.`, false, {
+  id: 'ember-cli-page-object.import-from-test-support',
+  until: "2.0.0",
+})
--- a/tests/integration/deprecations/test-support-test.js
+++ b/tests/integration/deprecations/test-support-test.js
@@ -1,0 +1,16 @@
+import { test, module } from 'qunit';
+import require from 'require';
+
+module('Deprecation | test-support');
+
+test('import from test-support leads to the deprecation', function(assert) {
+  // Currently addon-exports-test statically imports `dummy/tests/page-object`.
+  // That leads to pre-loading and caching of the `dummy/tests/page-object` module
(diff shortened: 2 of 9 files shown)


tests/integration/deprecations/test-support-test.js
```diff
@@ -0,0 +1,16 @@
...
+test('import from test-support leads to the deprecation', function(assert) {
+  // Currently addon-exports-test statically imports `dummy/tests/page-object`.
+  // That leads to pre-loading and caching of the `dummy/tests/page-object` module
+  // on the tests startup phase.Then if we tryto  access this module again
+  // w/o cleaning the cache we won't be able to catch a deprecation.
+  window.require.unsee('dummy/tests/page-object');
+
+  require('dummy/tests/page-object');
+
+  assert.expectDeprecation(`Importing from "test-support" is now deprecated. Just import directly from the "ember-cli-page-object" module instead.`);
```
magistrula: Could this assertion somehow live in `addon-exports-test`, so that once we remove support for importing from `test-support`, we only have to update that one test?

ro0gr: My intention was to remove the whole `tests/integration/deprecations` directory once we switch to 2.0. Actually I'd prefer to move `test-support` import tests from `addon-exports-tests` here but these tests are not easy to extract..

magistrula: Yeah, I like your idea of moving the `test-support` import tests here...that addresses my question about reducing the # of files that will need to be updated later. It's not really that big a deal, though, so I think we're fine either way.

test-support/page-object.js
```diff
@@ -75,3 +77,8 @@ export default {
 };
 
 export { buildSelector, findElementWithAssert, findElement, getContext, fullScope } from 'ember-cli-page-object';
+
+deprecate(`Importing from "test-support" is now deprecated. Just import directly from the "ember-cli-page-object" module instead.`, false, {
```
magistrula: This is a total unimportant nitpick, but suggest replacing `just` with `please` here; I think that's more conventional language for deprecation messages.

san650:
Thanks @ro0gr and @magistrula, you're amazing!!

PR #369, opened 2018-01-24
deprecate usage of comma separated selectors

ro0gr:
part of #359

### DIFF
--- a/tests/integration/deprecations/comma-separated-selector-test.js
+++ b/tests/integration/deprecations/comma-separated-selector-test.js
@@ -1,0 +1,55 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+
+import PageObject from 'dummy/tests/page-object';
+
+// In the old ember testing API we have a `render` method available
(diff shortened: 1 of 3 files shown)


tests/integration/deprecations/comma-separated-selector-test.js
```diff
@@ -0,0 +1,55 @@
...
+
+// In the old ember testing API we have a `render` method available
+// only in the `moduleForComponent` test helper. Unfortunatelly it requires
+// some component name to be passed as a first argument. That's why we pass
+// a `calculating-device` here despite the fact that we don't actually need it.
+moduleForComponent('calculating-device', 'Integration | comma separated selectors', {
+  integration: true
+});
+
+test('usage of comma-separated selector in the scope leads to a deprecation', function(assert) {
```
magistrula: Should we also be asserting that the deprecation does *not* show up when the selector does *not* contain a comma?

ro0gr: Thanks, good point.

san650:
@magistrula @ro0gr I've updated the PR with a link to a description of the deprecation. Let me know what you think.

PR #370, opened 2018-02-04
Use legacy collection in corner case

san650:
Also, link to migration guide in legacy collection warnings.

### DIFF
--- a/addon/-private/properties/collection.js
+++ b/addon/-private/properties/collection.js
@@ -114,14 +114,40 @@ import { collection as legacyCollection } from './collection/legacy';
  * @return {Descriptor}
  */
 export function collection(definition) {
-  if ('itemScope' in definition) {
+  if (useLegacyAPI(definition)) {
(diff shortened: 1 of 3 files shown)


addon/-private/properties/collection.js
```diff
@@ -114,14 +114,40 @@ import { collection as legacyCollection } from './collection/legacy';
...
+  //         }
+  //       })
+  //     });
+  //
+  // Although this doesn't make any sense and doesn't work in legacy
+  // collections definitions, using the new collection API in this cases break
+  // tests suites. For that reason we use the legacy definition and print a
+  // deprecation warning.
+  return ('itemScope' in definition) ||
+    ('scope' in definition && 'item' in definition);
```
ro0gr: I don't actually follow why do we care about `scope` here.

I think it's enough to make something like `('itemScope' in definition) || ('item' in definition)` ?

PR #374, opened 2018-02-04
Change collection signature

san650:
This PR changes the new collection's function signature:

before

```js
var page = create({
  users: collection({
    scope: 'table tr'
  })
});
```

after

```js
var page = create({
  users1: collection('table tr'),

  users2: collection('table tr', {
    text: text()
  })
});
```

The main reason for this is that we don't have an easy way to distinguish uses of the legacy collection API from the new collection API.

This PR also renders a warning for missing `itemScope` property in legacy collection API.

### DIFF
--- a/addon/-private/properties/collection.js
+++ b/addon/-private/properties/collection.js
@@ -1,4 +1,5 @@
 import { deprecate } from '@ember/application/deprecations';
+import { warn } from '@ember/debug';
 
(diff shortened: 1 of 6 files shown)


san650:
@ro0gr @magistrula @pzuraq let me know your thoughts about this change. Or alternatives to it.

PR #375, opened 2018-02-04
Enable array like accessors when Proxy is available

san650:


san650:
I'm experimenting with `Proxy` to enable array like access to elements in a collection. This uses progressive enhancement, so if you use a browser that supports it, you can use the feature in your tests.

What are your thoughts about something like this? I think it makes collections nicer to use than `.objectAt(...)`.

cc/ @magistrula @ro0gr @pzuraq

addon/-private/properties/collection/main.js
```diff
@@ -106,6 +106,22 @@ if (typeof (Symbol) !== 'undefined' && Symbol.iterator) {
   }
 }
 
+function proxyIt(instance) {
+  return new window.Proxy(instance, {
+    get: function(target, name) {
+      if (typeof(name) === 'number' || typeof(name) === 'string') {
```
san650: This is needed because `parseInt` blows up when called with a `Symbol` object (which we use to make it iterable).

pzuraq:
100% on board with this, this is great 😄 looking forward to being able to play around with it

PR #378, opened 2018-02-06
[Collection] Preserve behavior from legacy collection for non existing item

### LINKED ISSUE #377: New collection API requires andThen wrappers
Hello again! I love the new collection API, so much cleaner. In the course of updating the `travis-web` page objects, though, I found that I needed to add a bunch of `andThen`s that were previously unnecessary, whenever we were using a page object to perform an action.

For instance, in a test that cancels a build, with the old API, we have [this](https://github.com/travis-ci/travis-web/blob/293e373b85b162df56b33895c85398b8c05aee8a/tests/acceptance/repo/build-list-routes-test.js#L274):

```
  page.builds(0).cancelButton.click();
```

I mostly used regexes to update the uses of collections across the codebase, which changed it to this:

```
  page.builds[0].cancelButton.click();
```

However, the test was failing because that element wasn’t found! In all the places where we used page objects within a collection, I had to add an `andThen` wrapper; as you can see in [the commit](https://github.com/travis-ci/travis-web/pull/1516/commits/b54419030c7f18ccc151a956887360283fe56a06), there were quite a few instances.

I isolated the above example in [this commit](https://github.com/travis-ci/travis-web/pull/1516/commits/f34b4e3e26472f3b20b1d0f5d8c6596c60717199) to show [the failure on Travis](https://travis-ci.org/travis-ci/travis-web/builds/338112512#L571-L574). I also [tried](https://github.com/travis-ci/travis-web/pull/1516/commits/55f648ac1c5ef49326d36e4829a99c8749c2ec65) the non-proxy version, but the error persisted.

Is this intended behaviour? It seems like a pretty dramatic change.

ro0gr:
fixes #377 

@san650 @magistrula @pzuraq

### DIFF
--- a/addon/-private/properties/collection/main.js
+++ b/addon/-private/properties/collection/main.js
@@ -2,10 +2,6 @@
 import Ember from 'ember';
 import { buildSelector, assign } from '../../helpers';
 import { create } from '../../create';
-import {
-  ELEMENT_NOT_FOUND,
-  throwBetterError
-} from '../../better-errors';
 import { count } from '../count';
 import Ceibo from 'ceibo';
 
@@ -49,18 +45,7 @@ export class Collection {
       this._items[index] = tree;
     }
 
-    let itemPage = this._items[index];
-
-    if (index > this.length - 1) {
-      throwBetterError(
-        this.parent,
-        `${key}.objectAt(${index})`,
-        ELEMENT_NOT_FOUND,
-        { selector: itemPage.scope }
-      );
-    }
-
-    return itemPage;
+    return this._items[index];
   }
 
   filter(...args) {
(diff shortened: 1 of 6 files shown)


pzuraq:
👍 on this. Before we didn't have a clear advantage to one approach or the other, and in the future I once everyone converts to `async/await` in general I think we'll be able to bring back the assertion, but it seems like the error will be a pretty big burden until then.

backspace:
Thanks for doing this, @ro0gr; I tried it out in [this commit](https://github.com/travis-ci/travis-web/pull/1516/commits/f8190b3f163a5e2bf1e040d9a067ec993e5b6b05), where I removed the example `andThen` in #377, and the job passed! 🎉

ro0gr:
@backspace that's cool. Thanks for verifying.

Right now I'm sitting in the plane. Will return in two weeks. So in case if any work has to be done to get it merged, anybody feel free to pick this up.

san650:
Thanks everyone to help fixing this!

PR #382, opened 2018-02-21
Generate docs for v1.14

### LINKED ISSUE #381: Missing docs for new collection API
v1.14.x should be added to the docs.

see:
 - https://github.com/san650/ember-cli-page-object/issues/352
 - https://github.com/san650/ember-cli-page-object/pull/347

ro0gr:
Closes #381


magistrula:
@ro0gr This is so strange, but it looks like `create.md` was not generated. That has happened to me too, but it seems not to happen to @san650. You can just copy the old `create` file, since nothing has changed. 

Does/should this update involve any new files, or is it just new versions of existing files?

ro0gr:
mhm.. Seems like docs.js takes into account [properties dir only](https://github.com/san650/ember-cli-page-object/blob/b15e53c5f431309a917009c2a8d049a242f95c73/docs.js#L289). I'll take a look to fix this.

> Does/should this update involve any new files, or is it just new versions of existing files?

@pzuraq updated [collection](https://github.com/san650/ember-cli-page-object/blob/245e441728a3f35bdbad6dcd932500a20290fdeb/docs/v1.14.x/api/collection.md) docs to match the new collection API. However seems like we've lost legacy collection docs. The user still can switch the docs to v1.13.x to find the old collection docs but it does not seem very user friendly. 

I think the best option for now is to restore legacy collection docs at the first place and move new collection API docs to the subsection describing that it's experimental in the scope of v1.x and is going to be a default in the v2.x.

ro0gr:
I've updated PR with a create.md file generated with [updated](https://github.com/san650/ember-cli-page-object/pull/384) docs.js.

> However seems like we've lost legacy collection docs. 

After thinking about it more I couldn't come up with a good solution to include both new and legacy(probably should be placed in the migration guide section) collection API and not polluting the docs(mostly due to my imperfect prose skill I guess). So I'd prefer to leave an old collection API docs out of scope of this pr.

magistrula:
@ro0gr How about we add a link in the new collection docs to the `1.13.x` legacy collection docs?

san650:
I like the idea of just adding a link to the legacy collection docs.

ro0gr:
closing this, since It's replaced by https://github.com/san650/ember-cli-page-object/pull/391

PR #383, opened 2018-02-24
upgrade to ember-cli@3

ro0gr:


LICENSE.md
```diff
@@ -0,0 +1,9 @@
+The MIT License (MIT)
```
san650: Let's remove this one, we already have `LICENSE` file.

ro0gr: nice catch 👍

PR #385, opened 2018-03-05
Fix compatibility exports

bendemboski:
When using `require()`, we're not guaranteed that the module we're `requiring` is resolved yet, so it could just be an empty object with no exports. This has actually happened to a user, although it's tightly tied to their module dependency graph, and I haven't been able to easily repro it, nor do I think writing a unit test for it is feasible :( But the fix is to "lazy call" the `@ember/test-helpers` methods so we can be assured that they are resolved by the time we call them, even if they aren't when we export our compatibility functions.

### DIFF
--- a/addon/-private/compatibility.js
+++ b/addon/-private/compatibility.js
@@ -38,6 +38,7 @@
 const { require } = window;
 
 let helpers;
+let waitFn;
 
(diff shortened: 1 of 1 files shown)


ro0gr:
@bendemboski thanks!

PR #398, opened 2018-04-07
Upgrade `ember-cli-babel` version to use new modules syntax

esbanarango:
The current `ember-cli-babel` version `6.6.0` doesn't support the new modules syntax:

image 2018-04-07 at 12 56 08 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
   },
   "dependencies": {
     "ceibo": "~2.0.0",
-    "ember-cli-babel": "^6.6.0",
+    "ember-cli-babel": "6.12.0",
     "ember-cli-node-assets": "^0.2.2",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -35,7 +35,7 @@
   },
   "dependencies": {
     "ceibo": "~2.0.0",
-    "ember-cli-babel": "^6.6.0",
+    "ember-cli-babel": "6.12.0",
```
san650: Can you prefix the version with a caret? `^6.12.0`. Or do we need to pin this specific version?

san650:
Thanks for working on this issue @esbanarango

package.json
esbanarango: 👍

esbanarango:
Thank you for this addon @san650 🙏

PR #400, opened 2018-04-19
move to addon-test-support + working coverage reports

ro0gr:
This is the next iteration of https://github.com/san650/ember-cli-page-object/pull/399

Unfortunately we can't funnel the code which we want to include to the test-coverage reports. It is so because `ember-cli-code-coverage` [expects](https://github.com/kategengler/ember-cli-code-coverage/blob/77b5b9201060d634d281f64a5bb7b412773a8b89/index.js#L178) `addon-test-support` modules to have fixed module paths like `ember-cli-page-objects/test-support`. 

That's why `--re-export/` directory has been introduced. It's responsible for keeping one linear re-export modules, namely `index.js`, `extend.js` and `macros.js`. `--re-exports/` is funneled to the non-standard `/ember-cli-page-object` destination during the build. This allows us to leave public import paths unchanged and keep test coverage working at the same time.

There are few more notable changes in the first commit:
 - All the "public" modules are moved out of "-private/" dir upward
 - macroses moved from the "properties/" to "macros/" dir


ro0gr:
> I tried this fork with a large app, and was working fine.

@simonihmig I would highly appreciate if you could do it again against this branch :pray: 

@san650 do you feel comfortable with the new directory structure?
```
addon-test-support/
  --re-exports/
  -private/
  extend/
  macros/
  properties/
  create.js
  index.js

```

magistrula:
I tried running my large test suite against this branch, and I hit this error (we use node `4.8.4`):
```
The fs-updater package requires Node 6.0.0 or newer.
```

simonihmig:
@ro0gr I tried this now as well. It *seems* to work in general (I am using node 8 btw), but I am getting this error: 
```
Could not find module `ember-cli-page-object/-private/execution_context` imported from ...``` 
```

when importing 

```
import { getExecutionContext } from 'ember-cli-page-object/-private/execution_context';
```

Yeah, I know, it's private... 😬 But shouldn't this still work? 

*Re: importing `getExecutionContext`, I needed this (or thought I need it, happy to be told otherwise!) to make a custom page object helper (power-select) work with the new testing APIs, especially support `await` as well as chaining. But that's a different topic...*

ro0gr:
Great! Thank you for trying this out!

@magistrula seems like broccoli-merge-trees@3 updated minimal node version to be 6. Going to downgrade to broccoli-merge-trees@2 for now.

@simonihmig ya, currently importing of the private `getExecutionContext()` is the only option to implement a custom async page object property. I hope we'll come up with a public way to do it someday but I'd like to avoid this as a public module for now if possible.
I can see 2 possible ways to deal with that:
 - `getExecutionContext()` should be available under `ember-cli-page-object/test-helpers/-private/execution_context` module name. Maybe updating an import path from the app side is a viable compromise?.
 - I think we can re-export `getExecutionContext()` to preserve a module path with a deprecation... But then we are forced to develop a path forward for async props I believe.

@san650 @magistrula do you have opinions on the `getExecutionContext()` issue?

PR #401, opened 2018-05-08
Improve FastBoot compatibility

### LINKED ISSUE #397: FastBoot broken without jQuery
After removing jQuery, I get this in a app when it runs in FastBoot:

```
/Users/simonihmig/Projects/tlw/tmp/broccoli_persistent_filterautoprefixer_filter-output_path-ZtRjUHYd.tmp/assets/vendor.js:119501
	var el = document.createElement("fieldset");
	                  ^

TypeError: Cannot read property 'createElement' of undefined
    at assert (/Users/simonihmig/Projects/tlw/tmp/broccoli_persistent_filterautoprefixer_filter-output_path-ZtRjUHYd.tmp/assets/vendor/ecpo-jquery/dist/jquery.js:882:1)
    at /Users/simonihmig/Projects/tlw/tmp/broccoli_persistent_filterautoprefixer_filter-output_path-ZtRjUHYd.tmp/assets/vendor/ecpo-jquery/dist/jquery.js:2701:1
    at /Users/simonihmig/Projects/tlw/tmp/broccoli_persistent_filterautoprefixer_filter-output_path-ZtRjUHYd.tmp/assets/vendor/ecpo-jquery/dist/jquery.js:2752:1
    at /Users/simonihmig/Projects/tlw/tmp/broccoli_persistent_filterautoprefixer_filter-output_path-ZtRjUHYd.tmp/assets/vendor/ecpo-jquery/dist/jquery.js:36:1
    at /Users/simonihmig/Projects/tlw/tmp/broccoli_persistent_filterautoprefixer_filter-output_path-ZtRjUHYd.tmp/assets/vendor/ecpo-jquery/dist/jquery.js:40:1
    at ContextifyScript.Script.runInContext (vm.js:59:29)
    at VMSandbox.eval (/Users/simonihmig/Projects/tlw/node_modules/fastboot/src/vm-sandbox.js:14:16)
    at /Users/simonihmig/Projects/tlw/node_modules/fastboot/src/ember-app.js:156:15
   ...
```

When jQuery is missing, the own version shipping with this addon is imported here: 
https://github.com/san650/ember-cli-page-object/blob/17dbadd4e23310a9948a0767bff6ba395fc9c9ca/index.js#L28-L31

I am not sure why this is breaking FastBoot, while when Ember CLI imports jQuery (when it is not excluded by setting `vendorFiles`), this is not the case!? 🤔

On the other hand, I'm also wondering while jQuery is imported to `vendor.js`, while it is used only for tests. I think it could go into `test-support.js` instead, right?

I quickly tried this out, by adding `, { type: 'test' }` to the `import()` calls. And indeed this fixing the problem (as `test-support.js` is not used in FastBoot), and tests are still running fine. I can easily submit a PR if that seems fine!

The same applies not only to jQuery, but all of this addon's code: couldn't that go into `test-support.js` as well instead of `vendor.js`? There is no point in having the addon's code availabe in dev mode, only in test, or do I miss something. Changing that would be easy as well I think...

dfreeman:
This should fix #397. Newer versions of FastBoot don't set `process.env.EMBER_CLI_FASTBOOT`—instead there's only a single build step, and the runtime code is expected to be guarded by a `typeof FastBoot` check if it's not FastBoot-compatible.

I do think the change proposed in #399 is a good idea independent of this, but it looks like it got held up because of code coverage issues, and this solves the more immediate problem.

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,7 +15,10 @@ module.exports = {
         return {
           enabled: this._shouldIncludeFiles(),
           vendor: ['dist/jquery.js'],
-          destDir: 'ecpo-jquery'
+          destDir: 'ecpo-jquery',
+          processTree(tree) {
+            return require('fastboot-transform')(tree);
+          }
         }
       }
     }
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "ember-cli-babel": "^6.12.0",
     "ember-cli-node-assets": "^0.2.2",
     "ember-native-dom-helpers": "^0.5.3",
+    "fastboot-transform": "^0.1.3",
     "jquery": "^3.2.1",
     "rsvp": "^4.7.0"
   },
--- a/vendor/shims/ecpo-jquery.js
+++ b/vendor/shims/ecpo-jquery.js
@@ -1,23 +1,27 @@
-// Define an amd module called "-jquery".
-// It expects that we have already imported our own version of jQuery
-// which is going to be used in ember-cli-page-oject internally.
-//
-// This mode is used when we deal with `jquery`-less apps.
-(function() {
-  // prevent jquery provided by ember-cli-page-object
-  // to be set to `Ember.$` by Ember. 
-  var jq = self['$'].noConflict();
-  delete self['jQuery'];
+if (typeof FastBoot === 'undefined') {
+  // Define an amd module called "-jquery".
+  // It expects that we have already imported our own version of jQuery
+  // which is going to be used in ember-cli-page-oject internally.
+  //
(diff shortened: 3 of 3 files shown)


san650:
Thanks @dfreeman for the fix, I'll check the failing test from master.

ro0gr:
@dfreeman I've just merged https://github.com/san650/ember-cli-page-object/pull/400 which moves e-c-p-o to the "addon-test-support/". I think this PR is not needed anymore?

dfreeman:
Seems like it. Thanks @ro0gr

PR #405, opened 2018-05-18
Use Ceibo descriptor `get` for visitable

ro0gr:
~There is a single usage of Ceibo's `value` for only a `visitable` property.~ This is not true. the `value` is still [used](https://github.com/san650/ember-cli-page-object/blob/master/addon-test-support/properties/collection/main.js#L118) in the new Collection. But seems like it's for a reason.

The only difference between `value` and `get` is that `value` creates a [`writable` descriptor](https://github.com/san650/ceibo/blob/master/index.js#L46-L51) while `get` doesn't. I think `visitable` should be changed to leverage Ceibo's `get` since I can see a reason to use writable descriptor for it.

### DIFF
--- a/addon-test-support/properties/visitable.js
+++ b/addon-test-support/properties/visitable.js
@@ -93,17 +93,19 @@ export function visitable(path) {
   return {
     isDescriptor: true,
 
-    value(dynamicSegmentsAndQueryParams = {}) {
-      let executionContext = getExecutionContext(this);
+    get() {
+      return function(dynamicSegmentsAndQueryParams = {}) {
+        let executionContext = getExecutionContext(this);
 
-      return executionContext.runAsync((context) => {
-        let params = assign({}, dynamicSegmentsAndQueryParams);
-        let fullPath = fillInDynamicSegments(path, params);
+        return executionContext.runAsync((context) => {
+          let params = assign({}, dynamicSegmentsAndQueryParams);
+          let fullPath = fillInDynamicSegments(path, params);
 
-        fullPath = appendQueryParams(fullPath, params);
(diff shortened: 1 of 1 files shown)


san650:
@ro0gr 

> The only difference between value and get is that value creates a writable descriptor while get doesn't.

This is not exactly true, the difference is that with `get`, every time you access the property, a new function is created. On the other hand with `value` you always get the same value. Is not so much about writable/readonly although that difference is also true.

The initial idea in Ceibo about having value and get was the ability to intercept a call to a property (with `get`) or not to intercept the call and just return a fixed value (with `value`).

Hope this clarifies things a bit 😃

ro0gr:
Yes, that makes things more clear. Thanks!

I don't see much difference in using of `get` or `value`. Actually there is but maybe Page Object doesn't need both `get` and `value`? It makes sense to me to use a single syntax(`get`) from the `Ceibo` for POs creation in the sake of simplicity. @san650 does it make sense to you or are the any reasons to use a `value` for the `visitable`?

san650:
@ro0gr for Page Object it makes total sense to only use `get` and we get the benefit of having access to the property name when evaluating the getter, which is useful for more descriptive errors.

---

Leaving more info about when `value` is used just in case someone finds it useful to know.

The `value` option is used automatically by Ceibo when you assign things that are not descriptors to a page object.

```
var page = create({
  foo: function() {
    return "FOO";
  }
})
```

In this case we generate the `foo` property in the page object _object_  using `value`. You can see this here https://github.com/san650/ceibo/blob/a4a9ad944249d0b14d40248f04bc162e5ac024e8/index.js#L40-L54

PR #406, opened 2018-05-18
Use Ceibo descriptor `get` for new collection

ro0gr:
This is the last Ceibo's `value` usage mentioned in https://github.com/san650/ember-cli-page-object/pull/405


san650:
Can we rebase the PR on top of master? I think you already fixed the failing tests on master, right?

ro0gr:
Hm.. seems like it changes `collections` behaviour in an un-desirable way:

```js
const p = create({
  c: collection('.selector')
});

p.c[0] === p.c[0] // => false
```
prior to this change the nodes were equal.. Going to cover it by test and revert the commit

PR #409, opened 2018-06-11
[WIP] [doc] Re-write Quickstart

ro0gr:
I think updating the docs is a prerequisite for cutting a stable 1.15 release. 

Initially I've tried just to replace all the `andThen`s with `async`/`await` but the current Quickstart heavily depends on Acceptance/Integration APIs from the legacy `ember-test-helpers`.

The new Quickstart page tries to address several issues:

  - Refresh Ember test API to use `@ember/test-helpers` and `qunit-dom`
  - More focus on the general EmberCLI Page Object patterns:
    - Component-first page objects
    - Essential built-in props examples
    - Essential custom properties examples
    - Basic extensibility and composibility patterns

It also puts the legacy test API example to the very bottom of the page. I've done it this way just because leveraging of the new test APIs dramatically reduces the explanations complexity IMHO.

### DIFF
--- a/guides/quickstart.md
+++ b/guides/quickstart.md
@@ -4,43 +4,368 @@ title: Quickstart
 ---
 
 {% raw %}
-This is a short guide to get you started writing page objects and using them in your acceptance and integration tests.
+This is a short guide to get you started writing page objects and using them in your tests.
 
-- [Acceptance tests](#acceptance-tests)
-- [Integration tests](#integration-tests)
+- [The Challenge](#the-challenge)
+- [Modeling Components](#modeling-components)
+- [Macroses](#macroses)
+- [Higher Order Methods](#higher-order-methods)
+- [Application Tests](#application-tests)
+- [Interoperability](#interoperability)
 
-## Acceptance Tests
+## The Challenge
 
-Suppose we have a couple of acceptance tests to test the login page of our site.
+Suppose we have a simple login form component with the following result markup:
+
+```html
+<form>
+  <span data-test-error></span>
+
+  <div data-test-username class="has-error">
+    <label for="username">Username:</label>
+    <input id="username" />
+    <span class="error-message"></span>
+  </div>
+
+  <div data-test-password class="has-error">
+    <label for="password">Password:</label>
+    <input id="password" />
+    <span class="error-message"></span>
+  </div>
+
+  <button [data-test-save]>Save</button>
+</form>
+```
+
+First let's take a look on a typical Ember component tests for this form:
 
 ```js
-test('logs in sucessfully', function(assert) {
-  visit('/login');
-  fillIn('#username', 'admin');
-  fillIn('#password', 'secret');
-  click('button');
+test('it requires username and password', async function(assert) {
+  await render(hbs`{{login-form}}`);
 
-  andThen(function() {
-    assert.equal(currentURL(), '/private-page');
-  });
+  await click('[data-test-save]');
+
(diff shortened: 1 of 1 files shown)


guides/quickstart.md
```diff
@@ -4,426 +4,459 @@ title: Quickstart
...
 
-```bash
-$ ember generate page-object login
+  await click('[data-test-save]');
 
-installing
-  create tests/pages/login.js
-```
+  assert.dom('[data-test-username]').hasClass('has-error');
+  assert.dom('[data-test-username] .error-message').hasText('Username is required');
```
ro0gr: `[data-test-username] [data-test-error]`?

guides/quickstart.md
```diff
@@ -4,426 +4,459 @@ title: Quickstart
...
+  // ...
+}
 ```
 
-Let's use our existing page object to refactor these integration tests. As a reminder, here is our page object. (We don't need to change anything to use it in our integration tests!)
+That's it. Our `username` is ready to be used in the tests. But what about the `password`? Should we copy-paste all the `username` implementation into it?
+
+## Macroses
+
+Well, the only difference between  the `username` and `password` is a `scope` selector. It could be annoying to repeat the whole field definition accross all the similar fields.
```
ro0gr: Seems like `login-form` is too complex for the Quickstart page. I think user defined "macroses" better fits to the "Extensibility" page.

guides/quickstart.md
```diff
@@ -4,426 +4,459 @@ title: Quickstart
 ---
 
 {% raw %}
-This is a short guide to get you started writing page objects and using them in your acceptance and integration tests.
+This is a short guide to get you started writing page objects and using them in your tests.
 
-- [Acceptance tests](#acceptance-tests)
-- [Integration tests](#integration-tests)
+- [The Challenge](#the-challenge)
+- [Modeling Components](#modeling-components)
```
ro0gr: should be moved to the components docs

guides/quickstart.md
```diff
@@ -4,426 +4,459 @@ title: Quickstart
...
 });
 ```
 
-Let's take a look at the changes:
+Now all the DOM implementation details of the `login-form` are abstracted away with a Page Object. It improves test readability and feel safer while refactoring. 
 
-- In the test's `beforeEach()` hook we set the page's test context with `page.setContext(this)`. That tells the page object to use the test's `this.$()` to find elements, instead of Ember's global acceptance test helpers.
-- In the `afterEach()` hook, we call `page.removeContext()` to clear the test context from the page object.
-- The rest of the changes are the same as in our acceptance tests: After you set the test's `this` context on the page object, you can use the page object as before. (The one exception is `page.visit()`, which doesn't work in component tests since we don't have access to a router.)
+## Higher Order Methods
```
ro0gr: should be a part of the Components page

ro0gr:
After keeping it for a while I think the current `login-form` example should be changed to something other.
The main issues are:
 -  duplication of username/password fields requires macros to share definitions which is too complicated for the Quickstart I believe. 
 - A comprehensive form field requires a comprehensive Page Object definition. I think complex component definitions should be placed in on the Components page
 - Collections are not mentioned at all. It's quite difficult to me to find a natural use case for collections on the login page :laughing:

Currently I think something like a list/table with a search field may be a better option for the Quickstart.

ro0gr:
Closing in favour of https://github.com/san650/ember-cli-page-object/pull/415

PR #410, opened 2018-06-11
[doc]: use async/await in the components.md

ro0gr:
as a part of updating the docs to the new testing API

Background: Recently I've incidentally [updated](https://github.com/san650/ember-cli-page-object/commit/e268ef0fa625401a5accbc03ea0b01f3083f1d3f) an "Overview" page with a direct commit 
 to use `async`/`await`. It happened because of some confusion with GH interface. Now we basically can revert it or keep updating the docs. Personally I'd prefer the second option.

### DIFF
--- a/guides/components.md
+++ b/guides/components.md
@@ -47,16 +47,12 @@ const page = create({
   }
 });
 
-page
+await page
   .visit()
   .form
   .firstName('John')
   .lastName('Doe')
   .submit();
-
-andThen(function() {
-  // assert something
-});
 ```
 
 ## Default attributes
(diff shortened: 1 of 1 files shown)


magistrula:
@ro0gr Looks like there are some other uses of `andThen` that we could replace...do you want to handle all of them in this PR? I noticed some in `quickstart.md` & `as.js`,

magistrula:
Ah! Just noticed #409, where you are rewriting Quickstart.

ro0gr:
Closing in favour of #415.

PR #411, opened 2018-06-12
[feat] Allow native getter syntax

pzuraq:
This PR allows users to use native `get` syntax directly on page
objects, as opposed to the `getter` macro. This syntax is much less
verbose and easier to read overall.

The implementation first ensures that all property descriptors are
correctly copied to the final definition, then updates objects in the
Ceibo builder to use the equivalent of the getter macro. This could also
be added directly to Ceibo, but would likely still require some changes
to ember-cli-page-object to ensure that descriptors are copied via
`getOwnPropertyDescriptor` and `defineProperty` and not via simple
assignment (which will activate the getter when attempting to assign
it).

I'm also unsure where the best place to document this feature would be.
Should it just be a part of the API docs for `getter`?

### DIFF
--- a/guides/components.md
+++ b/guides/components.md
@@ -120,19 +120,24 @@ You can use native syntax for functions and getters on pages as well to create
 methods and properties with custom functionality:
 
 ```js
-import { click } from 'ember-test-helpers';
-import { findElement } from 'ember-cli-page-object';
+import { find, click } from 'ember-test-helpers';
 
 const page = create({
-  button: {
-    scope: 'button',
+  form: {
+    scope: '.awesome-form',
 
-    clickWith(options) {
-      click(findElement(this), options);
-    },
+    firstName: fillable('#firstName'),
+    lastName: fillable('#lastName'),
+    submit: clickable('button')
+
+    login(data) {
+      return this.username(data.username)
+        .password(data.password)
+        .submit();
+    }
 
     get role() {
-      return findElement(this).getAttribute('aria-role');
+      return find(this.scope).getAttribute('aria-role');
     }
   }
 });
--- a/guides/components.md
+++ b/guides/components.md
@@ -8,6 +8,7 @@ Group attributes and create new ones
 * [Components](#components)
 * [Default attributes](#default-attributes)
 * [Custom helper](#custom-helper)
+* [Functions and Getters](#scopes)
 * [Scopes](#scopes)
 
 ## Components
@@ -111,6 +112,30 @@ andThen(function() {
 });
 
 page.modal.clickOn("I'm sure");
+```
+
+## Functions and Getters
+
+You can use native syntax for functions and getters on pages as well to create
+methods and properties with custom functionality:
+
+```js
+import { click } from 'ember-test-helpers';
+import { findElement } from 'ember-cli-page-object';
+
+const page = create({
+  button: {
+    scope: 'button',
+
+    clickWith(options) {
+      click(findElement(this), options);
+    },
+
+    get role() {
+      return findElement(this).getAttribute('aria-role');
+    }
+  }
+});
 ```
 
 ## Scopes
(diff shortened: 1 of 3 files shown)


ro0gr:
Seems it can be documented as a part of regular components workflow.

guides/components.md
```diff
@@ -113,6 +114,30 @@ andThen(function() {
...
+```js
+import { click } from 'ember-test-helpers';
+import { findElement } from 'ember-cli-page-object';
+
+const page = create({
+  button: {
+    scope: 'button',
+
+    clickWith(options) {
+      click(findElement(this), options);
```
ro0gr: I think we can't recommend direct usage of external test helpers for actions. In order to implement a new type of action we require a private import `import { getExecutionContext } from 'ember-cli-page-object/-private/execution_context';` currently.

I believe we would [address](https://github.com/san650/ember-cli-page-object/commit/71adf959a35319196449f8cbea42bfb731c987ac) it soon. But for now we only support "read" functions and usages of actions defined on the page object context, like:
```
async fillIn(data) {
  await this.field1.fillIn('...');
  return await this.field2.fillIn('...');
}
```

ro0gr: What about?

```
login(data) {
  return this.username(data.username)
    .password('secret', data.password)
    .submit();
}
```

pzuraq: I don't think execution context has to be imported. 

If the user is using native context, they _know_ they can just use `ember-test-helpers` directly. If the user is using the old jquery context, they know they should be using jquery. We can address this in the guides, but using plain functions should be fine, we use them all the time in our tests.

The key here is that when using function syntax, you are _inherently_ making one off functions that are _not_ macros. For instance, see [these helpers in Ember Table](https://github.com/Addepar/ember-table/blob/master/addon-test-support/pages/ember-table.js#L62-L75);

pzuraq: Actually I think I misread that. Agree that we can change the example to the one you suggested, since it doesn't have the same issue. I still think it's fine for users to use `findElement` as long as they know what execution context their app will be running in, but we would need to document that.

ro0gr: Yeah, using of findElement is totally ok

guides/components.md
```diff
@@ -113,6 +114,35 @@ andThen(function() {
...
+    submit: clickable('button')
+
+    login(data) {
+      return this.username(data.username)
+        .password(data.password)
+        .submit();
+    }
+
+    get role() {
+      return find(this.scope).getAttribute('aria-role');
```
pzuraq: This would not always be valid, but I can't think of another example for a getter that doesn't use `findElement`. Open to suggestions

ro0gr: `findElement`/`findElementWithAssert` is a public API provided to find DOM elements in the context aware fashion(it uses a proper "find" function based on the currect execution context). I think we can safely use it.

Well there is actually an issue that `findElement` currently returns a `jQuery` selection object.. But that's unrelated to the PR.

magistrula:
@pzuraq Could you add a test validating that properties using native `get` can be aliased? Here's an [example](https://github.com/san650/ember-cli-page-object/blob/9b01f3b3d3b325941370d35827ca982c48038b3c/tests/unit/-private/properties/alias-test.js#L246) of such a test for `getter`.

guides/components.md
magistrula: Most of my getters compute values from existing page object properties, like this:
```js
get isFormEmpty() {
  return !this.firstName && !this.lastName;
}
```

ro0gr:
@pzuraq am I right saying that you have a full plate already and you don't plan to revive this PR anytime soon? 

By the way, thanks for this great idea! I think it can be a cool replacement for `getter`.

pzuraq:
Yeah that's correct, I'm pretty busy and don't think I'll have time anytime soon. Feel free to take over and reopen it, sorry I wasn't able to finish it up!

ro0gr:
No worries, I hope at some point it should find its way to the codebase. Meanwhile closing the PR. 

Thanks again!

Issue #412, opened 2018-06-12
Docs content issues


ro0gr:
I think documentation currently wants some content enhancements. Opening this to discuss and track the progress of the guides enhancements for v1. There are few issues I found:

- [x] Usage of the old testing API. 

I think switching to `async`/`await` is a prerequisite for a stable 1.15 release. Migration from `andThen`s and Integration tests would dramatically improve readability.  
Related activity: https://github.com/san650/ember-cli-page-object/commit/e268ef0fa625401a5accbc03ea0b01f3083f1d3f, https://github.com/san650/ember-cli-page-object/pull/410

- [x] Components - It got better now with the v1.15.x docs

I think components are underrepresented in the docs currently. There are few usages of actions/properties where a regular component is enough. Examples should be component first. That means no more promotion of:
```
  firstName: fillable('#firstName'),
  lastName: fillable('#lastName'),
  submit: clickable('button')
```
Each component is supplied with `fillIn` and `click` by default. So definition should contain 3 separate page object components. There is a work ~in progress~ with some ideas (https://github.com/san650/ember-cli-page-object/pull/409).

- [x] Collections - mentioned on the Quickstart page in v1.15.x docs

Collections currently missing a dedicated page in the Guides. Also we can probably do a better job on the [Quickstart](http://ember-cli-page-object.js.org/docs/v1.14.x/quickstart) page to explain collections.

- [ ] Extend

We currently just drop a raw API on this page. We should likely move `find` methods docs to the API section and provide the page with a basic extension patterns like

- [ ] Native Events Mode

Probably the best idea is to remove the page and deprecate the feature :laughing:
I believe this mode is not really needed anymore since the new testing API has appeared.

ro0gr:
closing it. It seems easier open a documentation PR rather than maintaining a non-actionable issue

PR #415, opened 2018-06-21
Quickstart rewrite

ro0gr:
This refreshes and simplifies the "Quickstart" to use new testing API. There is only a single example of `moduleForIntegration` in the very bottom of the page which was provided just to show a requirement to set and remove `context`.

There are 3 tests explained here:
 - one rendering test using components
 - one rendering test using collections
 - one application test which combines component and collection from the rendering tests 

@san650 @magistrula what do you think? Would appreciate your review.

Replaces https://github.com/san650/ember-cli-page-object/pull/409
Part of https://github.com/san650/ember-cli-page-object/issues/412

todos/questions: 
 - [ ] Haven't tested examples yet
 - [ ] Is it ok to reference the addon as "EmberCLI Page Object"?
 - [ ] Explore if there are any conflicts with the other existing Guide pages


guides/quickstart.md
```diff
@@ -4,426 +4,179 @@ title: Quickstart
...
 
-```js
-import {
-  create,
-  clickable,
-  fillable,
-  text,
-  visitable
-} from 'ember-cli-page-object';
+Suppose we have a very simple search form:
```
ro0gr: ~very~

guides/quickstart.md
```diff
@@ -4,426 +4,179 @@ title: Quickstart
...
-  password: fillable('#password'),
-  submit: clickable('button'),
-  error: text('.errors')
-});
+  <button>Search</button>
+</form>
 ```
 
-Now we include the page object in the test and replace the existing test helpers with the page object's methods and properties.
+The form also supports `has-error` class when user does submit with an empty search text.
```
ro0gr: unsure if it's nice enough explanation :smiley:

ro0gr:
Also re-written "components.md" page. There are still missing some basic examples, like macroses or getters(https://github.com/san650/ember-cli-page-object/pull/411).

Under macroses I mean a component definition builder like:

```js
function input(scope) {
  return {
    scope,
    fillIn: fillable('input'),
    value: value('input'),
    label: { scope: 'label'},
    err: { scope: '.error-message'}
  }
}
```

PR #416, opened 2018-06-21
[bugfix] Prevent is-rfc268-test to be treated as a test

ro0gr:
From Slack:
> is-rfc268-test.js:4 Uncaught TypeError: Object.defineProperty called on non-object

so this happens with 1.15 betas only when trying to import `useNativeEvents` from the `ember-cli-page-object/extend`. 

I don't fully understand the root cause of the issue, but here are my observations:
 - `extend/index.js` depends on `-private/execution_context`
 - which depends on `-private/is-rfc268-test` which causes an issue acording to error message
 - Somehow when `-private/is-rfc268-test` is referenced via this depency chain out of the test execution context(for example in the `tests/test-helper.js`) we get an error above. if I rename the `-private/is-rfc268-test` to avoid the "-test.js" suffix an error dissapears.

It could possible fix the other imports issues I have heard about earlier..

I think there are pretty low chances to test it on CI cause we need an external app/addon to include `ember-cli-page-object` for that.


mydea:
Any chance of getting a new release? Would be great :)

ro0gr:
@mydea thanks for heads up! Going to make new beta release early this week.

PR #418, opened 2018-07-01
Correct example typo

veelenga:
Example full text:

```js
// <span>super text</span>
// <span>regular text</span>

  spansContain: PageObject.contains('span', { multiple: true })
});

// all spans contain 'text'
assert.ok(page.spanContains('text'));

```

Link: http://ember-cli-page-object.js.org/docs/v1.4.x/api/predicates#contains


san650:
Thanks @veelenga

Issue #198, opened 2016-06-12
[Proposal] an ability to interact with dynamically rendered fields 


ro0gr:
### The Problem

As far as I can see currently page objects are supposed to work only with static page layouts. I mean you as a developer should define all the possible page fields at the page object's design time.   

Let's say you have a fieldset which contains some controls. Each time user changes some dropdown  completely new controls are re-rendered. [Here](https://ember-twiddle.com/51d46fcbca54bc0e784ef77a127f50a0) is an example of what am I talking about.

Now if we need to write tests which depend on several dropdown states I can see only two options:
- declare all the possible fields in the page object
- implement page objects for each of dropdown states which is almost the same as the first ption

both variants seems redundant. It requires lots of copy-paste and pollutes the page object:

``` js
create({
  field1State1: fillable('.control[name="field1"]'),
  field2State1: selectable('.control[name="field2"]'),
  field1State2: fillable('.control[name="field1"]'),
  field2State2: clickable('.control[name="field2"]'),
  field3State2: selectable('.control[name="field3"]')
})
```
### Proposal

I think we may need to provide a way for `selectors` not to be completely hard-coded in the page objects. It can contain something like selector template string which can be compiled when it accessed from the test. 

An ideal(from my optimistic perspective) api may be something like this:

``` js
const page = create({
  // I use "&" as a placeholder to be replaced with an actual value 
  // when template compilation happened. Probably it should be something another
  // but I like analogy with `less`|`sass` here ;) 
  templatedInput: fillable('.control[name="&"]'),
  templatedCheckbox: clickable('.control[name="&"]'),
  templatedDropdown: selectable('.control[name="&"]')
});

page.visit()
    .templatedInput('field1', 'some text')
    .templatedCheckbox('field2')
    .templatedDropdown('field3', 'option5')
```

It allows us to significantly reduce lines of code and makes page object much more flexible.
### Drawbacks

I suspect an implementation of example above may require lots of work and rewriting existing methods.
### Alternatives

I've [tried to play](https://github.com/ro0gr/ember-cli-page-object/commit/72194a86eed305b6cb03b748d1fe6e280362ae2a) with this idea and not touching an original action implementations. It looks like this:

``` js
const page = create({
  templatedInput: lazyProp(fillable, '.control[name="&"]'),
  templatedCheckbox: lazyProp(clickable, '.control[name="&"]'),
  templatedDropdown: lazyProp(selectable, '.control[name="&"]')
});
```

[...]

Issue #352, opened 2017-12-15
New Collection API RFC


pzuraq:
After revisiting #347 I realized that there's a lot going on in that PR and I wanted to clarify the design and intent behind it. The proposal is ultimately a major departure from the existing `collection` API and I think it would be good to discuss it in a semi-RFC like manner.

## Motivation

The current collection API is somewhat difficult to work with for two main reasons:

1. It breaks the mental model of page objects by forcing item definitions into a "scope" which doesn't exist.

    For most pages, users can follow assume that the tree structure of the page definition exactly matches the tree structure of the final page. With collections, this is not true:

    ```js 
    let page = create({
      items: collection({
        scope: 'table', 
        caption: 'caption', 
        itemScope: 'tr', 
        resetScope: true,

        item: {
          foo: 'bar'
        }
      })
    });

    page.items(0).foo; // 'bar' 
    page.items(0).caption; // undefined
    page.items().forEach((i) => i.foo); // 'bar' 
    page.items().forEach((i) => i.caption); // 'caption' 
    page.items().caption // 'caption'
    ```

    As you can see, there is no `item` scope for users to chain off of. Users must also know how to follow the various rules of where values that defined in collections end up:

    * The `scope` that is passed into the collection is not actually related to the items themselves, it is the _container_ of the items. 
    * You must define the item's scope on `itemScope`, not in the item definition itself 
    * You can access other values on the collection definition on the enumerable, and nowhere else (as in the caption example) 
    * `resetScope` must be set at the root collection level, and not the item level. Confusingly, it also applies to _both_ the items _and_ the item container. There is no way to specify `resetScope` for just one or the other.

    This is not a very complex set of rules, but it can be a bit counter-intuitive to newcomers to have to learn these details.

2. It operates as both a query function _and_ an enumerable depending on how users decide to use it, which puts us into different competing mental models. Users can either call the collection as a function to "query" it with a value, or call it without a value to receive back an enumerable and do more complex operations:

    ```js
    page.items(0); page.items().forEach(() => /**/); 
    ```

    This is confusing because we have no way of using the API consistently. Unlike jquery, queries return a single page without any of the collection functions (`$('.foo').each()` and `$('.foo:eq(0)').each()` are both valid, jquery always returns an iterable). The only way to access the enumerable is the special cases where no values are passed, and in turn this is the only way to access special properties like the `caption` from the previous example.

These API inconsistencies ultimately make it difficult to learn how collections work, and difficult to expand on them. 

[...]

PR #419, opened 2018-07-25
Add named collection items support

ro0gr:
Closes: #198, #288 

Also I think it should address most of use cases of the `findOne()` originally [proposed](https://github.com/san650/ember-cli-page-object/issues/352) by @pzuraq.

--------

This is a quick try to implement named collection items support described in https://github.com/san650/ember-cli-page-object/issues/288#issuecomment-407566187

There are two new pieces added to the public API:

- `id` is a component definition attribute which gives collections a hint how to identify items in the DOM.
We can use any getter attribute(`attribute`, `property`, `text`,...) provided by page object. It's also possible to use `alias` to find an `id` in the child components ~~or even custom functions to define more complex id rules~~(currently collection expects `id` to be a getter).

- `getById()` allows to find collection element by id. It's similar to an `objectAt`, but it works with a string name rather than position.

### Example
```html
<!-- Some dynamic fields form -->
<form>
  <div>
    <label for="id1">label 1</label>
    <input id="id1" name="field1" />
  </div>
  <div>
    <label for="id2">label 2</label>
    <input id="id2" name="field2" />
  </div>
</form>
```
```js
const p = create({
  scope: 'form',
  fields: collection('>div', {
    id: alias('input.name'),

    label: {
      scope: 'label'
    },

    input: {
      scope: 'input',
      name: attribute('name')
    }
  })
})

await p.fields['field1'].input.fillIn('value1');
await p.fields['field2'].input.fillIn('value2');

// we can even destruct fields by names
const { field1, field2 } = p.fields;

assert.equal(field1.label.label, 'label 1');
assert.equal(field2.label.text, 'label 2');
```

#### Naming

tbh I'm not very excited about `id` name. It can be confused with a DOM `id` attribute. Maybe `key` is a better name?

-----

**TODO**: 
 - [ ] update docs
 - [ ] handle cases when `id` is not defined 
 - [ ] decide on naming
 - [ ] more tests/edge cases

### DIFF
--- a/addon-test-support/properties/collection/main.js
+++ b/addon-test-support/properties/collection/main.js
@@ -48,6 +48,10 @@ export class Collection {
     return this._items[index];
   }
 
+  getById(id) {
+    return this.toArray().find(o => o.id === id);
+  }
+
   filter(...args) {
     return this.toArray().filter(...args);
   }
@@ -100,6 +104,10 @@ function proxyIt(instance) {
         if (!isNaN(index)) {
           return target.objectAt(index);
         }
+      }
+
+      if (typeof target[name] === 'undefined') {
+        return target.getById(name);
       }
 
       return target[name];
--- a/tests/integration/collection-id-test.js
+++ b/tests/integration/collection-id-test.js
@@ -1,37 +1,41 @@
 import { module, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
-import { render } from '@ember/test-helpers';
 import hbs from 'htmlbars-inline-precompile';
 import { create, collection, attribute } from 'ember-cli-page-object';
+import require from 'require';
 
-const page = create({
-  items: collection('ul>li', {
-    id: attribute('data-test-uuid')
-  })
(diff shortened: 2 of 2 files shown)


alexlafroscia:
A thought on naming -- rather than requiring that ID key be present, could you export a `Symbol` from the library that is used to identify what the "name" is for each collection item? That way you never have a collision between what should be used as a name and what is a property defined on the collection item.

It would look something like this:

```javascript
import { collectionItemName } from 'ember-cli-page-object';

const p = create({
  scope: 'form',
  fields: collection('> div', {
   [collectionItemName]: alias('input.name'),

    input: {
      scope: 'input',
      name: attribute('name')
    }
  })
})
```

ro0gr:
@alexlafroscia thank you for the hint. Using of `Symbol` looks interesting I will take it into consideration!

ro0gr:
I think a proper way to do it is to ressurect the `findOne` idea first. Then a keyed items access can be designed on top of it.

PR #420, opened 2018-07-31
[DOC] Document 'resetScope' for the Collection API

YoranBrondsema:
1. Documents 'resetScope' for the Collection API. The property still works in the new API when defined on the item definition but wasn't documented anywhere.
2. Remove `bower install` step in the README. There's no `bower.json` file anymore!


san650:
Thanks for contributing!

PR #431, opened 2018-11-18
[doc] Rewrite Quickstart page

ro0gr:
Another attemp to refresh the guides. This time I've focused on the "Quickstart" page only.

In order to explain things I've leveraged a calculator idea which I find a great example of PO usage.

I expect there may be some clumsy turns of speech in the proposed document. Hints to improve are highly appreciated!

My next immediate plan is to refresh a components page, which promises not to be such a huge rewrite like this is.

Part of https://github.com/san650/ember-cli-page-object/issues/412
Partially replaces https://github.com/san650/ember-cli-page-object/pull/415

### DIFF
--- a/guides/quickstart.md
+++ b/guides/quickstart.md
@@ -110,13 +110,13 @@ And, the same result with chaining:
 
 For comprehensive testing of our application, we must rely on more than testing individual components in isolation (ie acceptance testing). We can map each "page" in our app (route + template) to a `page-object`, composing various `page-object-component`s together to form a complete representation of page.
 
-Let's take a look at how we might generate a page-object.
+Suppose we have a search page in our application. Let's generate a page-object for it.
 
 ```
-$ ember generate page-object my-page
+$ ember generate page-object search-page
 
 installing
-  create tests/pages/my-page.js
+  create tests/pages/search-page.js
 ```
 
 The generator created a file inside the directory `/tests/pages`. Using this directory allows us to more easily distinguish pages from components, which are located under `/tests/pages/components/`.
@@ -138,17 +138,19 @@ You also might have noticed that rather than exporting a plain definition, we ex
 
 We can include any number of nested components, attributes, or methods in the definition, just as we did for `page-object-component`s.
 
-Suppose we have a calculator on the page:
+Let's update the page object as follows:
 
 ```js
 // my-app/tests/pages/my-page.js
 import { create, visitable } from 'ember-cli-page-object';
-import Calculator from 'my-app/tests/pages/components/quickstart-calculator';
+import SearchForm from 'my-app/tests/pages/components/search-form';
 
 export default create({
   visit: visitable('/'),
 
-  calculator: Calculator
+  search: SearchForm,
+
+  results: collection('ol li article')
 });
 ```
 
@@ -158,11 +160,15 @@ A simple application test using a `page-object` might look like:
 // my-app/tests/acceptance/my-page-test.js
 import myPage from 'my-app/tests/pages/my-page';
 
-module('My Page', // ...
-  test('it renders a calculator', async function(assert) {
+module('Search Page', // ...
+  test('it works', async function(assert) {
     await myPage.visit();
 
-    assert.ok(myPage.calculator.isVisible);
+    await myPage.search.text.fillIn('some');
+    await myPage.search.submit();
+
+    assert.equal(myPage.results.length, 1);
+    assert.ok(myPage.results[0].contains('Awesome search result!'));
   })
 ```
 
--- a/guides/quickstart.md
+++ b/guides/quickstart.md
@@ -12,7 +12,7 @@ This is a short guide to get you started building powerful UI testing APIs using
 
 ## Basic Component
 
-The primary build block of page objects is a [component](./components). It consists of a [`scope`](./components#scopes), [`attributes`](./components#attributes), custom methods and nested components.
+The primary build block of page objects is a [component](./components). It consists of a [`scope`](./components#scopes), [`attributes`](./components#attributes), custom methods, and nested components.
 
 Assume there is a simple `<SearchForm />` component with a text field and a submit button on it.
 
@@ -25,7 +25,7 @@ __Example__
 </form>
 ```
 
-Let's generate a page object component definition for it with a help of corresponding generator:
+Let's generate a page object component definition for it with the help of the corresponding generator:
 
 ```bash
 $ ember generate page-object-component search-form
@@ -108,7 +108,7 @@ And, the same result with chaining:
 
 ## Pages
 
-For comprehensive testing of our application, we must rely on more than testing individual components in isolation (ie acceptance testing). We can map each "page" in our app (route + template) to a `page-object`, composing various `page-object-component`s together to form a complete representation of page.
+For comprehensive testing of our application, we must rely on more than testing individual components in isolation (ie acceptance testing). We can map each "page" in our app (route + template) to a `page-object`, composing various `page-object-component`s together to form a complete representation of the page.
 
 Suppose we have a search page in our application. Let's generate a page-object for it.
 
@@ -121,7 +121,7 @@ installing
 
 The generator created a file inside the directory `/tests/pages`. Using this directory allows us to more easily distinguish pages from components, which are located under `/tests/pages/components/`.
 
-A generated page output looks like the following:
+The generated page object looks like the following:
 
 ```js
 // my-app/tests/pages/my-page.js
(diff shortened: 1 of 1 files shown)


guides/quickstart.md
```diff
@@ -4,426 +4,305 @@ title: Quickstart
...
-  enterName: fillable('input.username'),
-  register: clickable('button.register'),
-  myFanfare: MyFanfare
+
+  calculator: Calculator
 });
 ```
 
-Which will allow us to reference the `MyFanfare` component from the acceptance test.
+That's how a minimal application test using a page object can look like:
```
mistahenry: A simple application test using a `page-object` might look like:

mistahenry:
I think the component example with the form would greatly benefit from showing the markup of the form

juanazam:
I agree with all @mistahenry suggestions.

ro0gr:
@mistahenry this is an exact feedback I was looking for. Thanks a lot!

ro0gr:
> I think the component example with the form would greatly benefit from showing the markup of the form

Ya, I've also thought about it. As for me calculator markup takes to much of space on the page:

```html
<form class="quickstart-calculator">
  <span data-test-screen />

  <div class="numpad" >
    <button>1</button>
    <button>2</button>
    <button>3</button>
    <button>4</button>
    <button>5</button>
    <button>6</button>
    <button>7</button>
    <button>8</button>
    <button>9</button>
    <button>0</button>
  </div>

  <button class="plus">+</button>
  <button class="equals">=</button>
</form>
```

I believe we can get rid of such a wasteful template by specifying semantic tags on the PO definition:

```js
export default {
  scope: 'form.quickstart-calculator',

  plus: clickable('button.plus'),

  equals: triggerable('submit'),

  value: value('[data-test-screen]'),

  digits: collection('.numpad > button'),
};
```

Do you still think an explicit template can make documentation better?

---

A possible option might be using of `each` helper, which helps to reduce template size:

```hbs
<form class="quickstart-calculator">
  <span data-test-screen />

  <div class="numpad" >
    {{#each (array 1 2 3 4 5 6 7 8 9 0) as |num|}}
      <button>{{num}}</button>
    {{/each}}
  </div>

  <button class="plus">+</button>
  <button class="equals">=</button>
</form>
```

But then template gets a bit inconsistent cause it lacks the necessary Ember actions to be specified on the buttons and as a consequence the whole calculator js implementation(otherwise it's unclear why do we need actions at all) :)

In other words, until we have something like addon-docs integrated, I think we should avoid handlebars the documentation templates.

guides/quickstart.md
```diff
@@ -4,426 +4,305 @@ title: Quickstart
...
-  beforeEach() {
-    page.setContext(this);
-  },
-  afterEach() {
-    page.removeContext();
-  }
-});
+You also might have noticed that rather than exporting a plain definition we
+export a page object instance which can be used in your tests right after the page
+is imported.
```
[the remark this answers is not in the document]
ro0gr: Yes, this place probably needs an additional explanation. I think a reason for exporting a ready to use page object instances is that we don't need them to be extended.

> for performance?

Not sure what do you mean under performance..

mistahenry:
I really only meant show the form template for the very first component right before this:

```
import { create, triggerable } from 'ember-cli-page-object';

const myForm = create({
  scope: '.my-form',

  datum: { scope: '[data-test-datum]' },

  submit: triggerable('submit')
});
```

I think it's not necessary in the calculator example since the reader could extrapolate from the first example. And as you show above, it's way too much markup to digest in a `quickstart`. But, it may not hurt to casually link to [the dummy app calculator component](https://github.com/san650/ember-cli-page-object/blob/master/tests/dummy/app/templates/calculator.hbs)

guides/quickstart.md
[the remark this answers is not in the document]
mistahenry: I presumed the reason for advocating exporting page objects rather than definitions was for performance, since creating page object singletons once will be more performant than creating page objects per individual test case.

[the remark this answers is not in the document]
ro0gr: I believe, creating page object instances has very low impact on the test-suite performance. I don't believe we can slowdown a test-suite even if we create page object instances per each test run 😆 

I think that the reason for advocating for exporting of page objects is just a fact that a specific app page has a well known shape(set of components, URL, ...) and it rarely needs to be extended.

guides/quickstart.md
```diff
@@ -4,426 +4,305 @@ title: Quickstart
...
-
-export default create(MyFanfare);
+export default create({
+  visit: visitable('/')
+});
 ```
 
-This separation gives us two `import`-able signatures. In the case of the component's integration test importing the `default` will work as expected:
+The page is provided with a [`visitable`](./api/visitable) attribute to open
+associated web page.
```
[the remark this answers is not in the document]
ro0gr: I'm not sure about the 
> via URL/

when we define a `visitable`, we use a route, like `visitable('/page/:id')`. 

when we use it, we can only pass route params, like:
```
page.visit({id: 1});
```

Going to use your suggestion, but w/o `via URL/` part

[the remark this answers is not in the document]
mistahenry: The main issue was "to open associated web page." must be "to open **the** associated web page".

guides/quickstart.md
```diff
@@ -4,426 +4,278 @@ title: Quickstart
...
 ```
+$ ember generate page-object my-page
 
-And that's it! Our integration and acceptance tests are cleaner, more maintainable and easier to read.
+installing
+  create tests/pages/my-page.js
+```
 
-## Appendix
+The generator created a file inside the directory `/tests/pages`. It allows to easier distinguish pages between components, which are located under `/tests/pages/components/`.
```
mistahenry: "Using this directory allows us to more easily distinguish pages from components, which are located under `/tests/pages/components/`"

mistahenry:
> Going to:
> 
> incorporate the Basics section into the components guide

Having just reread through this page,  I think the calculator example has too much going on as the first thing you'd see when reading a quickstart. The short "Basics" example actually primes you for the calculator example. You explain the concepts more deeply in the components page so I didn't necessarily find the basic section and the components page to be *that* overlapping.

> rename Calculator section to smth like Building a Component
> I believe these changes would move us closer to a real quickstart.

In my opinion, the calculator section should replace the modal example in the components page. This would then be a quickstart with just basics (which is basically components) + pages. Very easy to digest at this point

guides/quickstart.md
```diff
@@ -4,426 +4,172 @@ title: Quickstart
...
-We want to convert these tests to use a page object.
+```html
+<form class="search-form">
+  <input type="search">
+  <button>Search</button>
+</form>
+```
 
-First, we need to create a new page object. For this we'll use one of the generators that comes with the addon.
+Let's generate a page object component definition for it with a help of corresponding generator:
```
mistahenry: Let's generate a page object component definition for it with **the** help of **the** following generator:

guides/quickstart.md
```diff
@@ -4,426 +4,172 @@ title: Quickstart
...
-  visit('/login');
-  fillIn('#username', 'admin');
-  fillIn('#password', 'secret');
-  click('button');
-
-  andThen(function() {
-    assert.equal(currentURL(), '/private-page');
-  });
-});
+The primary build block of page objects is a [component](./components). It consists of a [`scope`](./components#scopes), [`attributes`](./components#attributes), custom methods and nested components.
```
mistahenry: custom methods[,] and nested components.

guides/quickstart.md
```diff
@@ -4,426 +4,172 @@ title: Quickstart
...
+```
 
-  function submit(username, password) {
-    assert.equal(username, 'admin');
-    assert.equal(password, 'secret');
-  }
+## Pages
 
-  this.set('submit', submit);
+For comprehensive testing of our application, we must rely on more than testing individual components in isolation (ie acceptance testing). We can map each "page" in our app (route + template) to a `page-object`, composing various `page-object-component`s together to form a complete representation of page.
```
mistahenry: ...to form a complete representation of **the** page.

guides/quickstart.md
```diff
@@ -4,426 +4,172 @@ title: Quickstart
...
-And that's it! Our integration and acceptance tests are cleaner, more maintainable and easier to read.
+installing
+  create tests/pages/search-page.js
+```
 
-## Appendix
+The generator created a file inside the directory `/tests/pages`. Using this directory allows us to more easily distinguish pages from components, which are located under `/tests/pages/components/`.
 
-A helpful tip is to separate the exports in component page objects. This will allow you to compose larger page objects using the same definitions. For example say we have an integration test of a `my-fanfare` component:
+A generated page output looks like the following:
```
mistahenry: The generated page looks like the following:

mistahenry:
Seems good to go to me.

PR #432, opened 2018-11-19
feat: add helper for setting a page object in an integration test

alexlafroscia:
I've found myself using this helper in our codebase at work pretty frequently, and figured it might be useful to other people as well

The API is intended to mirror that of other QUnit helpers, where you pass the `hooks` object into the function to add setup/teardown logic.

This has been really helpful for cases where we have a page object for a component and want to use it within its integration tests as well as composing it into other page objects for acceptance tests.

### DIFF
--- a/addon-test-support/setup-page.js
+++ b/addon-test-support/setup-page.js
@@ -1,0 +1,11 @@
+import { create } from './create';
+
+/**
+ * Sets up a reference to `this.page` with the correct context set
+ */
+export default function(hooks, pageDefinition) {
+  hooks.beforeEach(function() {
+    this.page = create(pageDefinition);
+    this.page.setContext(this);
+  });
+}
--- a/addon-test-support/index.js
+++ b/addon-test-support/index.js
@@ -24,6 +24,7 @@ import { visitable }   from './properties/visitable';     export { visitable };
 export { findElement } from './extend/find-element';
 export { findElementWithAssert } from './extend/find-element-with-assert';
 export { buildSelector, getContext } from './-private/helpers';
+export { default as setupPage } from './setup-page';
 
 export default {
   attribute,
--- a/tests/integration/hooks-test.js
+++ b/tests/integration/hooks-test.js
@@ -1,54 +1,65 @@
-import { moduleForComponent, test } from 'ember-qunit';
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
 import { createCalculatorTemplate } from './test-helper';
 
-import PageObject from 'ember-cli-page-object';
+import PageObject, { setupPage } from 'ember-cli-page-object';
 
 const page = PageObject.create({});
 
-let firstThis;
-let secondThis;
+module('Integration | hooks', function(hooks) {
+  setupRenderingTest(hooks);
 
-moduleForComponent('calculating-device', 'Integration | hooks', {
-  integration: true,
+  module('manually setting up a page', function(hooks) {
+    let firstThis;
+    let secondThis;
 
-  beforeEach() {
-    page.setContext(this);
+    hooks.beforeEach(function() {
+      page.setContext(this);
 
-    if (!firstThis) {
-      firstThis = page.context;
-    } else {
-      secondThis = page.context;
-    }
-  },
+      if (!firstThis) {
+        firstThis = page.context;
+      } else {
+        secondThis = page.context;
+      }
+    });
 
-  afterEach() {
-    page.removeContext();
-  }
-});
+    hooks.afterEach(function() {
+      page.removeContext();
+    });
 
-test('When set in the `beforeEach()` qunit hook, test\'s `this` context\'s methods are accessible to the page object', function(assert) {
-  assert.expect(5);
+    test('When set in the `beforeEach()` qunit hook, test\'s `this` context\'s methods are accessible to the page object', function(assert) {
+      assert.expect(5);
 
-  assert.ok(page.context);
-  assert.deepEqual(this, page.context);
-  assert.equal(page.context, firstThis);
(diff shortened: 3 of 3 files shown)


addon-test-support/setup-page.js
```diff
@@ -0,0 +1,11 @@
+import { create } from './create';
+
+/**
+ * Sets up a reference to `this.page` with the correct context set
+ */
+export default function(hooks, page) {
+  hooks.beforeEach(function() {
+    this.page = create(page);
+    this.page.setContext(this);
```
ro0gr: What is the purpose of `setContext`? I think it's only applicable for the `moduleForIntegration` tests, but that doesn't seem a case here, cause we use use `setupTest` thing?

ro0gr: I think I've got it. You just avoid this annoying `setContext` thing.
I think it enforces all the pages which are instantiated with this helper to switch to an integration mode which makes it unusable in the other kind of tests.

alexlafroscia: We're currently only using this helper for integration tests, since acceptance tests don't need the page object to be created and the context set.

I was under the impression that it was required for you to use the `create` and `setContext` APIs when using page objects in integration tests, and this little helper just made the setup easier. Is that not the case?

alexlafroscia:
Could you elaborate more on what this would do?

```javascript
const page = this.lookupPage('some/page/object');
```

Where is it resolving that path relative to?

addon-test-support/setup-page.js
alexlafroscia: `moduleForIntegration` is the "old" version of the `ember-qunit` API -- `setupRenderingTest` is the "new" version of the same context, so using `setContext` on the page object is required, as far as I know.

ro0gr: Yes, you are right, `setContext` is required to enable page objects within `moduleForIntegration` tests.

However there is a [support of new test helpers](https://github.com/san650/ember-cli-page-object/pull/357) included in the beta of ember-cli-page-object. I you aware of that? Is there any  reason you want to keep using `moduleFor`?

ro0gr:
> Could you elaborate more on what this would do?

I think it could lookup a page object definition using `require`, and create a page object instance from it, smth like:

```js
function lookupPage(path) {
  const def = require.entries[`${appName}/tests/pages/${path}`].module.exports.default;

  return create(def);
}
```

But now, when I've written this down I realize this isn't really related to the `setupTest` thing cause it has nothing to do with a test context. It could be implemented just as a plain function.

I'm not actually sure if there any real benefits of the `lookupPage` approach. It probably should not pollute your PR, sorry for that!

alexlafroscia:
Gotcha. Additionally I wouldn't want to assume that page objects are inside `tests/pages` of the app or addon -- I've been putting them in `addon-test-support` for addons, so that apps that install the addon have the page objects in an easily import-able location that's only part of the `test` output.

ro0gr:
> I wouldn't want to assume that page objects are inside tests/pages

That's interesting. You've changed my mind 😃

addon-test-support/setup-page.js
ro0gr: As far as I understand, this way it's possible to use the latest test-helpers while keep using Ember jquery events.

There few issues with `setContext`:
- It's limited to legacy Ember integration mode only.
- It has to be set and removed for each page object instance.

I think we need a public way to explicitly specify mode the enable, smth like:

```js
import { setupPageObjects } from 'ember-cli-page-object/test-support';
import { EmberJqueryAdapter } from 'ember-cli-page-object/adapters'

module('...', function(hooks)
  setupPageObjects(hooks, {
    adapter: EmberJqueryAdapter
  });
```

Going to write some issues for this soon.

bendemboski:
My 2 cents -- I think the different test modes and how they relate to different page object execution contexts is already confusing enough that we shouldn't add to the confusion with something like this.

If I understand it, this allows you to use the legacy integration test context with QUnit nested modules. I don't think this is actually possible with `moduleForComponent()` tests -- I don't think they can give you access to the `hooks` that you would need to pass into the `setupPage()` method. So in order to use this, you'd have to use it with a `setupRenderingTest()` or `setupApplicationTest()` test, and then this would force `ember-cli-page-object` to use the execution context meant for `moduleForComponent()` tests.

While that would technically work, it would be using jQuery instead of `@ember/test-helpers`' native DOM helpers, moving the user squarely off of the intended/beaten path for Ember testing. So I don't think this is something we should add support for, because it's supporting an unintended/edge case, and is likely to add to the already-somewhat-problematic confusion about test styles and execution contexts.

alexlafroscia:
> I think the different test modes and how they relate to different page object execution contexts is already confusing enough that we shouldn't add to the confusion with something like this.

Totally agree.

I'll admit that I don't really know anything about execution contexts or when you need jQuery vs. not or why any of that is necessary. There is a really high likelihood that we're using this helper in our tests right now but should actually be doing something else.

bendemboski:
@alexlafroscia the TL;DR is that you want to call `page.setContext(this)` if and only if you're running a `moduleForComponent()` test. If you're running a `moduleForAcceptance()` test, a `setupRenderingTest()` test, or a `setupApplicationTest()`, e-c-p-o will detect it and do the right thing.

Calling `page.setContext(this)` forces e-c-p-o to use synchronous jQuery calls to interact with the DOM. If you're running a `setup*Test()` test, you don't really want that, you want it to auto-detect and use `@ember/test-helpers` helpers with native events and whatnot.

alexlafroscia:
Cool. In that case, I'm going to go and close this PR because it doesn't seem necessary. I'll look at removing this helper in my own app since we don't need it there, either!

PR #439, opened 2018-12-11
Upgrade to ember-cli@3.5

ro0gr:
Updates internals to use ember/ember-cli of v3.5.

There is also some cleanup included, which is related to extending ignore files rules with assets produced by `jekyll`, etc..

The issue with upgrade is that with the new blueprint for some of ember-try scenarios we have `jquery-integration` feature set to off which makes old-style global testing helpers non-working.

Going to keep the following strategy:
 - [x] skip `moduleFor` tests if `Ember.$` is not available
 - [x] re-write some of old tests from `moduleForComponent` to `rendering` tests
 - [x] re-enable jquery by default. This would allow to keep all the tests in the dev mode


ro0gr:
Seems like some changes in the travis config re-enabled double CI build. `travis-ci/push` was disabled before this PR, will investiagate.

PR #440, opened 2018-12-12
Do not re-create collection on consequent access

ro0gr:
see https://github.com/san650/ember-cli-page-object/pull/406#issuecomment-446722236

### DIFF
--- a/addon-test-support/properties/collection/main.js
+++ b/addon-test-support/properties/collection/main.js
@@ -115,8 +115,11 @@ export function collection(scope, definition) {
   let descriptor = {
     isDescriptor: true,
 
-    get(key) {
-      return proxyIfSupported(new Collection(scope, definition, this, key));
+    setup(node, key) {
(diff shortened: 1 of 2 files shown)


addon-test-support/properties/collection/main.js
```diff
@@ -115,8 +115,11 @@ export function collection(scope, definition) {
   let descriptor = {
     isDescriptor: true,
 
-    get(key) {
-      return proxyIfSupported(new Collection(scope, definition, this, key));
+    setup(_node, key) {
+      // Set the value on the descriptor so that it will be picked up and applied by Ceibo.
+      // This does mutate the descriptor, but because `setup` is always called before the
+      // value is assigned we are guaranteed to get a new, unique Collection instance each time.
+      descriptor.value = proxyIfSupported(new Collection(scope, definition, this, key));
```
mistahenry: this should be `descriptor.value = proxyIfSupported(new Collection(scope, definition, _node, key));`

Issue #444, opened 2019-02-06
ENHANCEMENT - `setupPageComponentTest` for hooks with easier setup with test APIs


rtablada:
Just a suggestion in our app we have started using a hook helper function (in QUnit tests) like this:

```
export function setupPageComponentTest (hooks, page) {
  hooks.beforeEach(function() {
    page.setContext(this);
  });

  hooks.afterEach(() => {
    page.removeContext();
  });
}
```

This removes a lot of boilerplate from our component tests since we don't need to set the context and remove context for component tests using EPO.

I'm not sure if this is something other teams would value?

The one worry I have (and why I did not create a PR yet) is that this helper is coupled to `ember-qunit`...

san650:
I think it makes sense to include it in the package. Can you create a PR?

rtablada:
Yeah I'll make a PR and some docs on how to use ECPO in integration tests for components.

ro0gr:
Hi, there is actually a PR already for that https://github.com/san650/ember-cli-page-object/pull/432

ro0gr:
As far as I understand this helper is supposed to enable `moduleForComponent` mode for the page object. Is it correct or do you need it in the `setupRenderingTest` for some reason?

I think a test helper for a specific page object is a bad path to go. What if the test needs few page objects to use? The relevant test setup would look something like:

```js
  setupPageComponentTest(hooks, page1);
  setupPageComponentTest(hooks, page2);
```
which doesn't look good to me, since we never want different test modes within the same test.

I mean that the test mode seems like a global state within a test to me. But we currently store and read it from the [specific pageObject instance](https://github.com/san650/ember-cli-page-object/blob/master/addon-test-support/-private/execution_context.js#L27). I think with the new test helper we can do better, by setting a test mode globally and then take care to clean it up in `afterEach`, so we don't need to pass a page object instance to it:

```js
  setupPageObjects(hooks, {
    moduleForComponent: true
  })
``` 
I think it's even possible to do it in a backward compat way(supporting `setContext` as a fallback)

Hope it makes sense and sorry for quite confusing thoughts dump(a bit under preassure now)...

bendemboski:
My 2 cents: https://github.com/san650/ember-cli-page-object/pull/432#issuecomment-461577885

PR #446, opened 2019-02-07
FIX - Root Element Lookup Not Context Lookup

rtablada:


rtablada:
When using `getElements` to reproduce the behavior of `@ember/test-helpers` then `findElements` needs to use `getRootElement` otherwise the context for acceptance tests is the inner HTML element and not the test container Ember application.

This causes issues with things like ember-modal-dialog and other top level wormholes that would usually be found by `find` and the Ember test helpers but are failing with ember-cli-page-object.

bendemboski:
I'm 👍 on the PR, but have a point of clarification. I believe that this will change the behavior of formerly-integration tests, while preserving the behavior of formerly-acceptance tests.

What I mean is that the legacy integration context uses the equivalent of [this.$(selector)](https://github.com/san650/ember-cli-page-object/blob/568074dc9fca9479b009e568073e5aa437b7e23f/addon-test-support/-private/execution_context/integration.js#L142) to find elements, so without specifying a custom `testContainer` integration tests _would not_ find content inside the root element but outside `this.element`. The legacy acceptance context uses the old `find()` helper, which searches inside the application's root element -- the equivalent of [$(selector, getRootElement())](https://github.com/emberjs/ember.js/blob/cafbae485d2459b95fcca4301f7a63ad395701d3/packages/ember-testing/lib/helpers/find.js#L38), so without specifying a custom `testContainer`, acceptance tests _would_ find content inside the root element but outside `this.element`.

With RFC232/268, we only have one execution context, and it currently searches inside `this.element`, which matches the behavior of the legacy integration test context. This PR changes it to search inside `getRootElement()`, which matches the legacy acceptance test context.

It has to differ in behavior from either the legacy acceptance test context or the legacy integration test context, and I think searching inside `getRootElement()` is the better way to do it, especially because that matches the behavior of [@ember/test-helpers](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/dom/-get-element.ts#L20), and I believe of `qunit-dom`.

But I think it's worth mentioning and considering that this will change the behavior of formerly-integration tests. I'm not sure if it's worth putting in some documentation somewhere, since the only test failures I could think of that this change would cause would be if a test were depending on _not_ finding an element inside the root but outside `this.element`...

rtablada:
For integration tests this should still work. If you set the context for the test using `page.setContext` then nothing will change with this PR since it only changes the behavior of the `else` when the context is not changed.

We have both acceptance and integration tests running with this so I will double check that everything is fine.

As far as compatibility with `setupApplicationTest` and `setupRenderingTest`, the `find`, `click` etc, from `@ember/test-helpers` are always pinned to `getRootElement`

bendemboski:
Oh sorry, I wasn't clear -- I meant across an update. So if you update an integration test from old-style to RFC232-style without changing your page objects, then they will behave different with this PR, whereas previously they would have behaved the same (and the reverse is true of updating acceptance tests).

Existing legacy integration tests won't change (although existing RFC232/268 tests _will_ change, but not in a way that is very likely to break anything.)

rtablada:
@bendemboski just checked.

We updated all of our integration tests to the 232 style and after this change there was not a behavior shift.

We are being explicit and setting the page context as outlined in this suggested enhancement: https://github.com/san650/ember-cli-page-object/issues/444

ro0gr:
LGTM, will test it in few apps tomorrow.

ro0gr:
Published as v1.15.2 🎉 

Thank you guys!

PR #447, opened 2019-02-28
Extract run

ro0gr:
An original motivation for this change was a desire to learn of how do actions+chaining work here per each context. As a result knowledge about about chainability was extracted from execution contexts was extracted to a standalone `run` function. 

For the sake of keeping a backward compatibility in legacy contexts(moduleFor) a new `andThen` hook was added. For contexts that have `andThen`, `run` waiting for the previous action to be completed before invoking a new one. I believe we should remove `andThen` in the next major release.

### DIFF
--- a/addon-test-support/-private/action.js
+++ b/addon-test-support/-private/action.js
@@ -1,0 +1,97 @@
+import { resolve } from 'rsvp';
+import { getExecutionContext } from './execution_context';
+import { getRoot } from './helpers';
+import Ceibo from 'ceibo';
+
+/**
+ * Run action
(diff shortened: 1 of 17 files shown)


ro0gr:
CI should be fixed., when new ember-qunit is published with https://github.com/emberjs/ember-qunit/pull/547

ro0gr:
Recently some native events tests started to fail in master. I was surprised that chainability change fixes it, probably because we take more control over async behavior with switching to `run`.

Also there are some other fresh test failure fixes in this PR.

PR #448, opened 2019-03-01
Deprecate setContext

### LINKED ISSUE #444: ENHANCEMENT - `setupPageComponentTest` for hooks with easier setup with test APIs

ro0gr:
Fixes #444 

`moduleForComponent` + `setContext` was a point  of much confusion for ages. With @ember/test-helpers we are able to auto-detect if it's `moduleForComponent` kind of test or not.

### DIFF
--- a/addon-test-support/-private/context.js
+++ b/addon-test-support/-private/context.js
@@ -54,6 +54,12 @@ export function render(template) {
  * @return {PageObject} - the page object
  */
 export function setContext(context) {
+  deprecate('setContext() is deprecated. Please make sure you use "@ember/test-helpers" of v1 or higher.', false, {
+    id: 'ember-cli-page-object.set-context',
+    until: '2.0.0',
+    url: 'https://ember-cli-page-object.js.org/docs/v1.16.x/deprecations/#set-context',
+  });
+
   if (context) {
     this.context = context;
   }
--- a/addon-test-support/-private/execution_context.js
+++ b/addon-test-support/-private/execution_context.js
@@ -1,5 +1,5 @@
 import { getContext as getIntegrationTestContext } from './helpers';
-import { getContext, visit } from './compatibility';
+import { getContext as getEmberTestHelpersContext, visit } from './compatibility';
 import AcceptanceExecutionContext from './execution_context/acceptance';
 import IntegrationExecutionContext from './execution_context/integration';
 import Rfc268Context from './execution_context/rfc268';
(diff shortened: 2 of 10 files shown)


addon-test-support/-private/execution_context.js
```diff
@@ -24,56 +24,22 @@ export function getExecutionContext(pageObjectNode) {
   // first, and only if that hasn't happened, check to see if we're in an
   // RFC232/RFC268 test, and if not, fall back on assuming a pre-RFC268
   // acceptance test, which is the only remaining supported scenario.
-  let integrationTestContext = getIntegrationTestContext(pageObjectNode);
+  let context = getTestContext() || getRoot(node).context;
+
+  const isModuleForIntegration = typeof context === 'object' && context !== null
+    && typeof context.$ === 'function'
+    && !context.element;
```
ro0gr: `context.element` exists in the `setup*Test`'s context. When the `element` is not present it's a `moduleFor*`

addon-test-support/-private/execution_context.js
```diff
@@ -1,5 +1,5 @@
-import { getContext as getIntegrationTestContext } from './helpers';
-import { getContext, visit } from './compatibility';
+import { getRoot } from './helpers';
+import { getContext as getTestContext } from './compatibility';
```
ro0gr: surprisingly, `getContext` is able to return a `moduleFor` test context. We don't need to always store it on the page object.

addon-test-support/-private/context.js
```diff
@@ -54,6 +54,12 @@ export function render(template) {
  * @return {PageObject} - the page object
  */
 export function setContext(context) {
+  deprecate('setContext() is deprecated. Please make sure you use "@ember/test-helpers" of v1 or higher.', false, {
+    id: 'ember-cli-page-object.set-context',
+    until: '2.0.0',
+    url: 'Fill Me',
```
ro0gr: needs a new deprecations page in the guides

addon-test-support/-private/execution_context.js
```diff
@@ -28,10 +28,14 @@ export function getExecutionContext(pageObjectNode) {
...
-    contextName = 'rfc268';
-  } else {
+  } else if (isAcceptanceTest()) {
     contextName = 'acceptance';
+  } else if (supportsRfc268()) {
+    contextName = 'rfc268';
+  }
+
+  if (!contextName) {
+    throw new Error('Can not detect test type. Please make sure you use the latest version of "@ember/test-helpers".');
```
ro0gr: needs coverage

addon-test-support/-private/context.js
ro0gr: waiting for https://github.com/san650/ember-cli-page-object/pull/465

addon-test-support/-private/execution_context.js
ro0gr: unfortunatelly, in added test, there are some `require.unsee` conflicts with https://github.com/san650/ember-cli-page-object/blob/master/tests/unit/-private/compatibility-test.js, which I could not figure out.
Removing a new test :(

PR #451, opened 2019-04-08
empty create regression

kellyselden:
Somewhere between 1.3.0 and 1.5.2, an empty create started failing.

### DIFF
--- a/tests/integration/default-properties-test.js
+++ b/tests/integration/default-properties-test.js
@@ -76,5 +76,13 @@ if (require.has('@ember/test-helpers')) {
       assert.equal(page.dummy.isVisible(), 'isVisible');
       assert.equal(page.dummy.text(), 'text');
     });
+
+    test('allows empty create', async function(assert) {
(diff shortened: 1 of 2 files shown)


tests/integration/default-properties-test.js
```diff
@@ -76,5 +76,14 @@ if (require.has('@ember/test-helpers')) {
       assert.equal(page.dummy.isVisible(), 'isVisible');
       assert.equal(page.dummy.text(), 'text');
     });
+
+    test('allows empty create', async function(assert) {
+      let page = create();
+      page.setContext(this);
```
ro0gr: I think `setContext` should be removed, cause it only needed for `moduleForComponent` test

kellyselden: removed.

ro0gr:
Thanks for the fix! Will merge and publish this Sunday

ro0gr:
Just published as 1.15.3. Thanks!

PR #458, opened 2019-09-02
feat: Add ambient types

ro0gr:
It is a result of few iterations and testing against a real APP test suite.
It covers most of the current public APIs and provides you with auto-complete or errors when you mistype something.

Some of currently public APIs are intentionally ignored. For example, the current implementation of `create` supports the following signatures:

```js
  function create(definition) {}
  function create(definition, options) {}
  // not supported
  function create(url, definition) {}
  function create(url, definition, options) {}
```
I [believe](https://github.com/san650/ember-cli-page-object/issues/464), `url` in `create` should be avoided in favor of `visitable` action, so typings from this PR does only support 2 first signatures. This approach allows us to simplify typings and signal a user about a recommended way of doing stuff. 

Here is a list of all public API, which are currently exposed to the consumers, but don't have types:
 - `buildSelector`
 - `getContext`
 - `setContext`
 - `is` 

### Limitations

There are some hints for those who decide to write their page objects in TS.

#### alias

Typescript isn't able to follow nested string paths, like `a.path.to.some.deep.property`, so each `alias` returns `any`:

```ts
const page = create({
  doSomething: alias('nestedObject.click'),
  nestedObject: {}
});

page.doSomething(); // returns `any`
```

#### Custom methods

Because page object definitions are POJOs, which are static instances, in custom methods we can't have a valid `this` type. So for custom methods and getters we have to set `this` to `any`:

```ts
export default {
  something: getter(function(this: any) {
    return this.text as string; // but you can still provide a hint about your return value!
  })
}
``` 

the same is true for custom actions:

```ts
const page = create({
  doSomething(this: any) { 
    return this.click();
  }
})

page.doSomething(); // returns `any`
```

#### Documentation

It's missing because I didn't want to duplicate it by copy-pasting it from the JS source code. Someday I hope we would have a truly typed codebase, which would solve the documentation issue, but today I had few stoppers for doing this:
 - `create` - not sure if it's possible to type underlying `Ceibo`
 - code coverage doesn't work with ec-typescript@2 - https://github.com/kategengler/ember-cli-code-coverage/issues/213

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ LICENSE.md
 /docs
 /_site
 /yarn-error.log
+jsconfig.json
 
 # IDE/editor files
 /.idea/
--- a/.npmignore
+++ b/.npmignore
@@ -27,6 +27,9 @@
 /yarn.lock
 /package-lock.json
 .gitkeep
+/types/*
+!/types/index.d.ts
+/guides
 
 # ember-try
 /.node_modules.ember-try/
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ jobs:
       script:
         - npm run lint:hbs
         - npm run lint:js
+        - npm run lint:ts
         - npm test
 
     # we recommend new addons test the current and previous LTS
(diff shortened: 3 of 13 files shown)


ro0gr:
A sidenote: I believe most of the `this: any` issues should go away, once we figure out a class syntax for definitions.

NullVoxPopuli:
Yaaaaaas.gif

Issue #453, opened 2019-04-10
Custom properties


yratanov:
Hi,

I am using Ember 3.3 + `"ember-cli-page-object": "^1.15.2"`.

I'm trying to add new property to be used in page object like this:

```
// page 

export default create({
  role: powerSelect('.role'),
  submit: clickOnText('.btn'),
});

// powerSelect property defined like this

export function powerSelect(selector) {
  return {
    isDescriptor: true,

    get(_) {
      return function(text) {
        wait().then(() => {
          selectChoose(selector, text);
        });

        return this;
      };
    }
  };
}

//test 
module('Some module', function(hooks) {
  setupApplicationTest(hooks);

  test('adding priority', async assert => {
    await visit('/');
    await page.role('Admin');
    await page.submit();
  });
});
```

But it doesn't work properly. `selectChoose` does not finish, while submit is already clicked.

I looked in the code and it seems that my helper return value should be wrapped in `runAsync` from `rfs268.js`, but it is a private api....

So in genearal I have these questions:
1. Why creation of new properties is not covered in the docs? Is it an unexpeted usage of the library?
2. Why `runAsync` for all contexts is private?
3. If I suppose to be able to write custom properties with interactions, how should I do it?

juanazam:
@yratanov you could this without defining a property

```
export default create({
  role: powerSelect('.role'),
  submit: clickOnText('.btn'),
});

function powerSelect(selector) {
  return function(text) {
    selectChoose(selector, text);
  });
}

```

If you are still feeling strong about using a property, I would import `getExecutionContext`, I know it's private, but we should move it to the public space so building properties is easier.

yratanov:
@juanazam But it will break chaining, isn't it?

juanazam:
As show it will, but:

```
export default create({
  role: powerSelect('.role'),
  submit: clickOnText('.btn'),
});

function powerSelect(selector) {
  return function(text) {
    selectChoose(selector, text);
    return this
  });
}
```
And using `async` `await` properly, it shouldn't.

yratanov:
@juanazam not sure what you mean by "properly". I tried as you showed in your example, and it doesn't wait.

juanazam:
@yratanov I don't have time to do it right now and test, but I'm pretty sure it's doable using await on the `selectChoose` helper.

yratanov:
@juanazam 

OK, so here what helped:

```js
[...]

PR #460, opened 2019-10-06
[experiment] Explore user actions

ro0gr:
Opened this PR to share the status and bootstrap a discussion about a desirable API to define user actions.

Fixes https://github.com/san650/ember-cli-page-object/issues/453

---

description tbd


ro0gr:
just re-written all the built-in actions with a new proposed `run`+`action`. It allowed to drastically cleanup execution contexts and (I believe) make actions implementation a bit more straightforward.

with this change all the execution context methods intended to manipulate with DOM elements do only accept an `element` instead of a combo of `selector`, `container`, `options`. 
Which means, a single invocation of `MyExecutionContext.click` would only interact with a single DOM node.

ro0gr:
this led me to https://github.com/san650/ember-cli-page-object/pull/509

PR #461, opened 2019-10-18
adds #findOne and #findOneBy methods for `collection`

yratanov:


ro0gr:
Thanks for the PR @yratanov! 

Yes, I think currently we miss an easy way to find a single elemement for collections. There were some discussions in this area already. When new collections API was developed, @pzuraq raised an idea of `findOne`:
 -New Collections API - https://github.com/san650/ember-cli-page-object/issues/352
 - `findOne` possible design - https://github.com/san650/ember-cli-page-object/pull/347#discussion_r160061630

I feel, in scope of page object collections, it may make sense to provide a bit more clever form of `find`, which finds exactly one item or throws error if only 0 more than 1 items was found.

What do you think? does it make sense for you?

yratanov:
@ro0gr I've added asserts for:
> finds exactly one item or throws error if only 0 more than 1 items was found.

Not sure about `findOne` though. All other methods are native, so it will be confusing having `filterBy` that accepts `field, value` and `findOne`/`findAll` with completely different API. They should all use same arguments IMO.

What do you think?

ro0gr:
`findAll` was supposed to behave the same as `filter` does, but the idea was to give it a power to accept query objects, like:

```js
  myCollection.findAll({
    someProp: 'some text'
  })
```

As for me, even if it seems a bit more convenient, this query objects are quite limited:
 - what about nesting
 - "greater(less) than" or anything more advanced  than equality check
 - larger api surface

In other words, in my opinion, `findAll` is not a necessary part of collections, since `filter` is more flexible and well known for the most of the devs already. 

`findOne` on the other hand, is a bit different from `find` for a good reason, it allows developer to get a quick feedback if something goes wrong. If we let `find` throw, it could confuse a developer, cause that's not how regular `find` works. That's why I think highlighting the new behavior via a different name is a good idea.

ro0gr:
Also I'm not sure about the `findBy`... 
I understand we already have `filterBy` and `mapBy`, but:
 - afaik there are no such methods in native JS arrays. These are methods which are added by Ember to the Array prototype. For instance, if someone installs [ember-disable-prototype-extensions](https://github.com/ember-cli/ember-disable-prototype-extensions) in their app or addon our `mapBy`/`filterBy` would not work.
 - After we got arrow functions it does not really make more convenience to me to use `***By(key, value)` extensions:

```js
myList.findBy('id', 3);
// vs
myList.find(i => i.id === 3);
```  
maybe that's just a matter of taste or I miss something, but in my opinion, a single `findOne`(or something like this) would be sufficient.

yratanov:
> In other words, in my opinion, findAll is not a necessary part of collections, since filter is more flexible and well known for the most of the devs already.

👍 

> findOne on the other hand, is a bit different from find for a good reason, it allows developer to get a quick feedback if something goes wrong. If we let find throw, it could confuse a developer, cause that's not how regular find works. That's why I think highlighting the new behavior via a different name is a good idea.

I don't think assert is such a big deal so that we need to type three more characters. `find`/`findBy` are also instantly recognisable, so users will know what args to pass without looking into docs.

> After we got arrow functions it does not really make more convenience to me to use ***By(key, value) extensions:

I use `***By` methods a lot and find it more convenient... I think it's pretty common in Ember to know/use these methods.

ro0gr:
> I don't think assert is such a big deal so that we need to type three more characters. 

Sorry me for being that guy, but I can't agree. Though `find` and `findOne` may feel similar, they are different. It introduces 2 new edge cases. 

With `findOne` we:
 - throw if no elements found. I suspect I could accept it mentally under the `find` name, if it were the only case, but
 - throw when there are multiple matching elements on the page. On the other hand, selecting the first element from many is completelly fine for a native `find`. I think this is even less intuitive than throwing when no elements found.

Because of these ^ edge cases, I think `findOne` can't be considered [substitutable](https://en.wikipedia.org/wiki/Liskov_substitution_principle#Principle) with `find`.

> so users will know what args to pass without looking into docs.

We can improve learnability by fisnishing [typings](https://github.com/san650/ember-cli-page-object/pull/458) effort. Also `findOne` and `find` have the same interface but exceptions.

Hope it makes sense to you.

yratanov:
@ro0gr ok, then what do we do? I really would like to keep ability to find by `property, value`. 
`findOne` and `findOneBy`?

ro0gr:
@yratanov yes. let's do it, if that's fine for you!

/cc @san650

yratanov:
@ro0gr renamed :)

addon-test-support/properties/collection/legacy.js
```diff
@@ -5,7 +5,7 @@ import { create } from '../../create';
 import { count } from '../count';
 import Ceibo from 'ceibo';
 
-const arrayDelegateMethods = ['map', 'filter', 'mapBy', 'filterBy', 'forEach'];
+const arrayDelegateMethods = ['map', 'filter', 'mapBy', 'filterBy', 'findOneBy', 'findOne', 'forEach'];
```
ro0gr: I suspect it does not not work for legacy collection cause it delegated to non-existing methods of the native Array? Maybe we should just take legacy collection alone?

tests/unit/-private/properties/collection-test.js
```diff
@@ -406,6 +407,137 @@ moduleForProperty('collection', function(test) {
...
+    });
+
+    await this.adapter.createTemplate(this, page, `
+      <span>Lorem</span>
+      <span>Ipsum</span>
+      <span>Ipsum</span>
+    `);
+
+    const expectedError = new EmberError(
+      `2 elements found by function, but expected 1\n\nPageObject: 'page.foo'`
```
ro0gr: I think now I understand why you like `findOneBy` variant more than `findOne`.. 

What do you think about changing `function` with `condition`? like:

> 2 elements found by **condition**

it's still generic, but looks a little bit better to my eye

yratanov:
@ro0gr applied your feedback 👍

ro0gr:
Thanks!

ro0gr:
@yratanov sorry for asking so late, but would you mind to add yourself to the [AUTHORS](https://github.com/san650/ember-cli-page-object/blob/master/AUTHORS) please?

addon-test-support/properties/collection.js
```diff
@@ -20,6 +20,8 @@ import { collection as legacyCollection } from './collection/legacy';
  * - `forEach()` - Runs a function for each item in the collection
  * - `map()` - maps over the elements of the collection
  * - `mapBy()` - maps over the elements of the collecton by the specified property
+ * - `find()` - finds first item of the array with assert by specified function
+ * - `findBy()` - finds first item of the array with assert by property
```
sukima, line 24: These are incorrect. This PR implements `findOne` and `findOneBy` but the docs here say it is `find` and `findBy`.

Also, Ember devs are used to arrays providing a `find` and `findBy` so it was astonishing to find this PR where the names were changed to `findOne` and `findOneBy`.

sukima: For reference because this is hidden and differed from common interfaces I'm used to I will be using this syntax instead: `[...page.myCollection].findBy('foo', 'bar')`

ro0gr: @sukima I'm sorry that you've faced that documentation issue. The docs should be fixed for sure.

> Also, Ember devs are used to arrays providing a find and findBy so it was astonishing to find this PR where the names were changed to findOne and findOneBy.

these methods have `One` for a reason. In comparison with a standard `find(` it does not look just for the first item in a list, but it ensures that the found item is the only one matching the filter condition, which is more determenistic and useful in tests.

Issue #463, opened 2019-10-25
Introduce DOM element finders


ro0gr:
## Summary

Introduce `findOne` and `findMany` for searching plain DOM elements by JQuery selector.

Deprecate existing `findElement`/`findElementWithAssert` in v1.

## Motivation

Today we have the following methods to find in the DOM:

```ts
function findElement(pageObject: PageObjectNode, scope?: string, options?: FindOptions): JQuery;

function findElementWithAssert(pageObject: PageObjectNode, scope?: string, options?: FindOptions): JQuery;
```

both of them return `jQuery` element collections which should be eliminated to reduce leakage of `jQuery` to the end user.

Also though they return collection, a collection contains a single Element by default. You can disable this behavior by passing an additional `multiple` option set to `true`. But here it starts conflicting with a finder name, cause with `multiple` we look for many element**s** with a  `findElement`.

## Detailed design

```ts
/**
 * Looks for a single element by JQuery selector
 *
 * @throws If no elements found
 * @throws If more than one element found
 */
function findOne(pageObject: PageObjectNode, scope?: string, options?: FindOptions): Element;

/**
 * Looks for elements matching by JQuery selector
 */
function findMany(pageObject: PageObjectNode, scope?: string, options?: FindOptions): Array<Element>;
```

Neither of these methods need a `multiple` flag, cause it's pretty straighforward which method to use when you need elements list. Also, If you just need to check if the element exists without throwing, you can check elements count returned by `findMany`. 

And probably the most important, we don't return a `JQuery` collection anymore. Note that we still allow jquery-like selectors, this is necessary to make `collection` work. Anyway, with the new finders we should be able to switch to a different search implementations compatible with jquery-like selectors.

## How we teach this

 - update docs
 - add a codemod to ease migration - [ember-page-object-codemod](https://www.npmjs.com/package/ember-page-object-codemod)

## Drawbacks

 It's a breaking change. However if we ever want to get rid of JQuery dependency we should avoid exposing its instances to the user, and that would be a breaking change anyways. 

## Alternatives

 - open to ideas about a better names for new finder methods
 - can we reduce number of args by putting `scope` to the `options`, so new interface gets:
```ts
// when no options passed
function findOne(pageObject: PageObjectNode, scope?: string): Element;
// scope is passed as a part of `options`
function findOne(pageObject: PageObjectNode, options?: FindOptions): Element;
```

yratanov:
[...]

PR #466, opened 2019-10-27
adds new dom finders 

yratanov:
#463

### DIFF
--- a/addon-test-support/extend/find-element-with-assert.js
+++ b/addon-test-support/extend/find-element-with-assert.js
@@ -39,7 +39,8 @@ import { deprecate } from '@ember/application/deprecations';
 export function findElementWithAssert(pageObjectNode, targetSelector, options = {}) {
   deprecate('findElementWithAssert is deprecated, please use findOne or findMany instead', false, {
     id: 'ember-cli-page-object.old-finders',
-    until: '2.0.0'
+    until: '2.0.0',
+    url: 'https://ember-cli-page-object.js.org/docs/v1.16.x/deprecations/#old-finders'
   });
   return getExecutionContext(pageObjectNode).findWithAssert(targetSelector, options);
 }
--- a/addon-test-support/extend/find-element-with-assert.js
+++ b/addon-test-support/extend/find-element-with-assert.js
@@ -1,4 +1,5 @@
 import { getExecutionContext } from '../-private/execution_context';
+import { deprecate } from '@ember/application/deprecations';
 
 /**
  * @public
@@ -36,5 +37,9 @@ import { getExecutionContext } from '../-private/execution_context';
  * @throws Will throw an error if multiple elements are matched by selector and multiple option is not set
  */
 export function findElementWithAssert(pageObjectNode, targetSelector, options = {}) {
+  deprecate('findElementWithAssert is deprecated, please use findOne or findMany instead', false, {
+    id: 'ember-cli-page-object.old-finders',
+    until: '2.0.0'
+  });
   return getExecutionContext(pageObjectNode).findWithAssert(targetSelector, options);
 }
--- a/addon-test-support/extend/find-element.js
+++ b/addon-test-support/extend/find-element.js
@@ -37,7 +37,8 @@ import { deprecate } from '@ember/application/deprecations';
 export function findElement(pageObjectNode, targetSelector, options = {}) {
   deprecate('findElement is deprecated, please use findOne or findMany instead', false, {
     id: 'ember-cli-page-object.old-finders',
-    until: '2.0.0'
+    until: '2.0.0',
+    url: 'https://ember-cli-page-object.js.org/docs/v1.16.x/deprecations/#old-finders'
   });
 
   return getExecutionContext(pageObjectNode).find(targetSelector, options);
--- a/addon-test-support/extend/find-many.js
+++ b/addon-test-support/extend/find-many.js
@@ -1,5 +1,5 @@
 import { getExecutionContext } from '../-private/execution_context';
-import { assertOptionsWhitelisted } from '../-private/helpers';
+import { filterWhitelistedOption } from '../-private/helpers';
 /**
  * @public
  *
@@ -30,7 +30,7 @@ import { assertOptionsWhitelisted } from '../-private/helpers';
  * @return {Array} of Element
  */
 export function findMany(pageObjectNode, targetSelector, options = {}) {
-  assertOptionsWhitelisted(options, ['resetScope', 'visible', 'testContainer', 'contains', 'scope']);
-  const opts = Object.assign({}, options, { multiple: true });
+  const filteredOptions = filterWhitelistedOption(options, ['resetScope', 'visible', 'testContainer', 'contains', 'scope']);
+  const opts = Object.assign({}, filteredOptions, { multiple: true });
   return getExecutionContext(pageObjectNode).find(targetSelector, opts).get();
 }
--- a/addon-test-support/extend/find-many.js
+++ b/addon-test-support/extend/find-many.js
@@ -1,5 +1,5 @@
 import { getExecutionContext } from '../-private/execution_context';
-
+import { assertOptionsWhitelisted } from '../-private/helpers';
 /**
  * @public
  *
@@ -24,13 +24,13 @@ import { getExecutionContext } from '../-private/execution_context';
  * @param {Object} options - Additional options
  * @param {boolean} options.resetScope - Do not use inherited scope
  * @param {string} options.contains - Filter by using :contains('foo') pseudo-class
- * @param {number} options.at - Filter by index using :eq(x) pseudo-class
- * @param {boolean} options.last - Filter by using :last pseudo-class
+ * @param {string} options.scope
  * @param {boolean} options.visible - Filter by using :visible pseudo-class
  * @param {string} options.testContainer - Context where to search elements in the DOM
  * @return {Array} of Element
  */
 export function findMany(pageObjectNode, targetSelector, options = {}) {
+  assertOptionsWhitelisted(options, ['resetScope', 'visible', 'testContainer', 'contains', 'scope']);
   const opts = Object.assign({}, options, { multiple: true });
   return getExecutionContext(pageObjectNode).find(targetSelector, opts).get();
 }
--- a/addon-test-support/extend/find-many.js
+++ b/addon-test-support/extend/find-many.js
@@ -25,12 +25,16 @@ import { filterWhitelistedOption } from '../-private/helpers';
  * @param {boolean} options.resetScope - Do not use inherited scope
  * @param {string} options.contains - Filter by using :contains('foo') pseudo-class
  * @param {string} options.scope
+ * @param {number} options.at - Filter by index using :eq(x) pseudo-class
+ * @param {boolean} options.last - Filter by using :last pseudo-class
  * @param {boolean} options.visible - Filter by using :visible pseudo-class
  * @param {string} options.testContainer - Context where to search elements in the DOM
  * @return {Array} of Element
  */
 export function findMany(pageObjectNode, targetSelector, options = {}) {
-  const filteredOptions = filterWhitelistedOption(options, ['resetScope', 'visible', 'testContainer', 'contains', 'scope']);
+  const filteredOptions = filterWhitelistedOption(options, [
+    'resetScope', 'visible', 'testContainer', 'contains', 'scope', 'at', 'last'
+  ]);
   const opts = Object.assign({}, filteredOptions, { multiple: true });
   return getExecutionContext(pageObjectNode).find(targetSelector, opts).get();
 }
--- a/addon-test-support/extend/find-many.js
+++ b/addon-test-support/extend/find-many.js
@@ -1,0 +1,36 @@
+import { getExecutionContext } from '../-private/execution_context';
+
+/**
+ * @public
+ *
+ * Returns array of elements
+ *
+ * @example
+ *
+ * import { findMany } from '../extend';
+ *
+ * export default function count(selector, options = {}) {
+ *   return {
+ *     isDescriptor: true,
+ *
+ *     get() {
+ *       return findMany(this, selector, options).length;
+ *     }
+ *   };
+ * }
+ *
+ * @param {Ceibo} pageObjectNode - Node of the tree
+ * @param {string} targetSelector - Specific CSS selector
+ * @param {Object} options - Additional options
+ * @param {boolean} options.resetScope - Do not use inherited scope
+ * @param {string} options.contains - Filter by using :contains('foo') pseudo-class
+ * @param {number} options.at - Filter by index using :eq(x) pseudo-class
+ * @param {boolean} options.last - Filter by using :last pseudo-class
+ * @param {boolean} options.visible - Filter by using :visible pseudo-class
+ * @param {string} options.testContainer - Context where to search elements in the DOM
+ * @return {Array} of Element
+ */
+export function findMany(pageObjectNode, targetSelector, options = {}) {
+  const opts = Object.assign({}, options, { multiple: true });
+  return getExecutionContext(pageObjectNode).find(targetSelector, opts).get();
+}
--- a/addon-test-support/properties/attribute.js
+++ b/addon-test-support/properties/attribute.js
@@ -1,6 +1,5 @@
 import { assign } from '../-private/helpers';
 import { findMany, findOne } from '../extend';
-import { A } from '@ember/array';
 
 /**
  * @public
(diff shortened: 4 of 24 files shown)


addon-test-support/properties/attribute.js
```diff
@@ -82,9 +83,11 @@ export function attribute(attributeName, selector, userOptions = {}) {
     get(key) {
       let options = assign({ pageObjectKey: key }, userOptions);
 
-      let elements = findElementWithAssert(this, selector, options);
-
-      return map(elements, element => element.attr(attributeName), options);
+      if (options.multiple) {
```
ro0gr: I know this makes total sense for backward compatibility. However..

I believe we should discourage users from passing `multiple` options for attributes. 
Such usages can/should be expressed with collections IMO, which would:
 - reduce API surface
 - simplify interanals, so each propoerty is always associated with a single node
 - lead to a more consistent page object definitions

so in my opinion, as the next step we should deprecate this behavior.

addon-test-support/extend/find-element.js
```diff
@@ -34,5 +35,10 @@ import { getExecutionContext } from '../-private/execution_context';
  * @throws Will throw an error if multiple elements are matched by selector and multiple option is not set
  */
 export function findElement(pageObjectNode, targetSelector, options = {}) {
+  deprecate('findElement is deprecated, please use findOne or findMany instead', false, {
+    id: 'ember-cli-page-object.old-finders',
+    until: '2.0.0'
+  });
```
ro0gr: will need to add a deprecation section and `url` once https://github.com/san650/ember-cli-page-object/pull/465 is merged

addon-test-support/extend/find-element-with-assert.js
```diff
@@ -36,5 +37,9 @@ import { getExecutionContext } from '../-private/execution_context';
  * @throws Will throw an error if multiple elements are matched by selector and multiple option is not set
  */
 export function findElementWithAssert(pageObjectNode, targetSelector, options = {}) {
+  deprecate('findElementWithAssert is deprecated, please use findOne or findMany instead', false, {
+    id: 'ember-cli-page-object.old-finders',
+    until: '2.0.0'
+  });
```
ro0gr: will need to add a deprecation section and url once #465 is merged

addon-test-support/properties/attribute.js
yratanov: 💯 May be we should add a deprecation message whenever it's used?

ro0gr: Yes, but I think we should keep it separate from this PR

ro0gr:
can we have it finders tests please?

I know it's tested implicitly via props, but once we start moving find responsibility from execution contexts to finders, I'd like to have it covered directly instead of working with non-obvious side-effect test failures.

yratanov:
@ro0gr I added basic tests, but not sure about `options` though.
`resetScope` - it seems like a sign of a bad architecture, breaks "page is a tree" concept.
`contains`, `at`, `last` - these will make tests brittle.
`testContainer` - don't know where it will be used. It seems to work in `Integration` and `Acceptance` tests without passing it to this function.

ro0gr:
I think all of the mentioned options make sense in certain situations.

> resetScope

`resetScope` is needed when you want to test elements which semantically belong to a parent element but rendered out of a normal flow. 
For example, let's consider a custom select element. It's usual that after you click on it you would see a dropdown which in fact rendered somewhere in a location(at the root of you `testContainer`) which doesn't belong to select's DOM element. Sometimes it's also possible to experience when you deal with modals, tooltips and whatnot..

> `contains`, `at`, `last` - these will make tests brittle.

I believe something like the following should work good enough.

```js
  const node = create();
  
  await render(hbs`<ul><li>item 1</li><li>item 2</li><li>item 3</li></ul>`)

  assert.equal(findOne(node, 'li', { contains: '2' }).innerText.trim(), 'item 2');
  assert.equal(findOne(node, 'li', { at: 2 }).innerText.trim(), 'item 2');
  assert.equal(findOne(node, 'li', { last: true }).innerText.trim(), 'item 3');
```

can't get what the issue is. Can you ellaborate please?

> testContainer

There are many contents around you test's viewport which is mostly QUnit stuff, and we want to scope our search. So by default it's `#ember-testing`. However, some libraries/addons might do some special things and render its DOM bound to `body` or some other unconventional locations.
It happens rare enough, but in this situation `testContainer` is very helpful!

tests/unit/extend/find-many-test.js
```diff
@@ -0,0 +1,50 @@
+import { module, test } from 'qunit';
+import { setupApplicationTest } from 'ember-qunit';
+import ApplicationAdapter from "../../helpers/properties/application-adapter";
+import { create } from 'ember-cli-page-object';
+import { findMany } from 'ember-cli-page-object/extend';
+import { findAll } from '@ember/test-helpers';
```
ro0gr: > Could not find module `@ember/test-helpers` imported from `dummy/tests/unit/extend/find-many-test`

fails for configuration w/o `@ember/test-helpers` installed. you can follow this pattern: https://github.com/san650/ember-cli-page-object/blob/629ddd3787b875567a011ad490805d348e9c2811/tests/integration/deprecations/comma-separated-selector-test.js#L7-L8

to only run the tests when `@ember/test-helpers` present

tests/unit/extend/find-many-test.js
```diff
@@ -0,0 +1,50 @@
...
+
+module(`Extend | findMany`, function(hooks) {
+  setupApplicationTest(hooks);
+
+  let adapter = new ApplicationAdapter(hooks);
+
+  test('finds by selector and returns array of Elements', async function(assert) {
+    let page = create({});
+
+    await adapter.createTemplate(this, page, '<em class="lorem"></em><span class="ipsum"></span>');
```
ro0gr: since at the end we should not care if about an actual execution context used, maybe we can use regular Ember testing APIs: `setupRenderingTest`, `await render`...?

addon-test-support/extend/find-element.js
ro0gr: Merged! 

can you please add old finders to the new deprecation page and add an `url` into here?

yratanov:
@ro0gr 

> assert.equal(findOne(node, 'li', { at: 2 }).innerText.trim(), 'item 2');
  assert.equal(findOne(node, 'li', { last: true }).innerText.trim(), 'item 3');

1) collections can be used.
2) If order was changed, test will be broken, users shouldn't rely on index.

ro0gr:
ah... I see your point. If we decide they are bad then I think we should not support it for new finders at all. 

However...

> collections can be used.

that's true, and that's what we do actually [use](https://github.com/san650/ember-cli-page-object/blob/629ddd3787b875567a011ad490805d348e9c2811/addon-test-support/properties/collection/main.js#L35) to make collections work.

> if order was changed, test will be broken, 

true as well, but we [need](https://github.com/san650/ember-cli-page-object/blob/629ddd3787b875567a011ad490805d348e9c2811/addon-test-support/properties/click-on-text/helpers.js#L11) `last` to make work `clickOnText`.

> users shouldn't rely on index.

I'm not 100% sure about that. It's pretty usual to me to use `myCollection[0]`, `myCollection[1]`,.. in my tests. Until my test data is stable, I can say "if order has changed, it's probably some issue in my code".

---

I agree that eaither of these options can be misused in userland. From the other hand if we as users wanted to implement custom `collections` or `clickOnText`, for example, we couldn't do it w/o such finder options. That makes me think we should leave them and treat as a public API.

yratanov:
@ro0gr It's more like a discussion about future APIs, rather than current implementation details. I believe collection implementation will be changed for no-jquery version in the future, so I thought we might think about the need of having jquery-like selector options.

> From the other hand if we as users wanted to implement custom collections or clickOnText

I think we can provide some basic collection API outside (if it's needed?). 
`clickOnText` seems to be very ambigous, imo we shouldn't promote this approach in testing as it can break easily on a small change.

yratanov:
@ro0gr not sure why coveralls is complaining...

ro0gr:
> not sure why coveralls is complaining...

I guess it's because you've removed a previously covered 'map' function. Fine for me

ro0gr:
> I think we can provide some basic collection API outside (if it's needed?).

JQuery's `last` is bit different. We need it to find the closest parent of a text:

```html
<div id="a">
  <div id="b">
    <div id="c">Test</div>
  </div>
</div>
```
```js
 $('div:contains("Test"):last').id === 'c' 
```

addon-test-support/extend/find-many.js
```diff
@@ -0,0 +1,36 @@
...
+ * @param {string} options.contains - Filter by using :contains('foo') pseudo-class
+ * @param {number} options.at - Filter by index using :eq(x) pseudo-class
+ * @param {boolean} options.last - Filter by using :last pseudo-class
+ * @param {boolean} options.visible - Filter by using :visible pseudo-class
+ * @param {string} options.testContainer - Context where to search elements in the DOM
+ * @return {Array} of Element
+ */
+export function findMany(pageObjectNode, targetSelector, options = {}) {
+  const opts = Object.assign({}, options, { multiple: true });
+  return getExecutionContext(pageObjectNode).find(targetSelector, opts).get();
```
ro0gr: it looks like we still support `multiple`(which is not intended) and the rest of the props that are supported by `getExecutionContext(pageObjectNode).find()`.
I think we should take it under control by picking and passing only the sane options + test each of them.

ro0gr:
Sorry I might have misunderstood your last comment. Do I get it correct that you propose to not expose some of the options to the user? 
My initial idea was to provide a very close alternative to existing finders so migration is smothier. But now I'm curious what would be your minimal set of options needed?

yratanov:
@ro0gr I would leave only `resetScope` and `testContainer`. I believe everythimg else can be achieved using query selectors, etc.

ro0gr:
> I would leave only resetScope and testContainer

I'm afraid that would break all the existing attributes, cause unfortunatelly they accept exactly the same options as the finders do:

```js
  {
    myAttr: attribute('my-attribute', {
      // any finder option is supported here
    })
  }
```

So we have to provide a clear alternative for each option which we are going to deprecate and then remove. That was not a point of the original issue #463. 

I'm not sure if it's worth efforts to fight with all the existing options at this point. TBH I'd be glad if we can figure out a path forward at least for `multiple`, cause I think it's more like a design flow  complicating internals, not only an anti-pattern. And I believe we have a good alternative for that: `Collections` and `findMany`.

Also, I'd like to try new finders in the [actions PR](https://github.com/san650/ember-cli-page-object/pull/460/files#diff-15bc2edbf7cd9275753f7c8af4c2a713R5). And as I've said earlier, by removing `contains`/`last`,...(?), we would make it more difficult to achieve.

That being said, I'd vote for leaving everything(but `multiple`) in place in order to keep changes incremental. Does it make sense for you?

addon-test-support/extend/find-many.js
```diff
@@ -0,0 +1,36 @@
...
+ * @param {boolean} options.resetScope - Do not use inherited scope
+ * @param {string} options.contains - Filter by using :contains('foo') pseudo-class
+ * @param {number} options.at - Filter by index using :eq(x) pseudo-class
+ * @param {boolean} options.last - Filter by using :last pseudo-class
+ * @param {boolean} options.visible - Filter by using :visible pseudo-class
+ * @param {string} options.testContainer - Context where to search elements in the DOM
+ * @return {Array} of Element
+ */
+export function findMany(pageObjectNode, targetSelector, options = {}) {
+  const opts = Object.assign({}, options, { multiple: true });
```
ro0gr: I've just checked which options are comming here... hope I haven't missed any:

```js
  testContainer,
  resetScope,

  // needed for collections
  at,

  // needed for clickOnText
  contains,
  last,

  // not documented
  pageObjectKey,
  scope,

  // hm...
  visible
```

I think we should explicitly pick all the known `options` here and test each of them.

yratanov: 👍  will add

ro0gr:
> So we have to provide a clear alternative for each option which we are going to deprecate and then remove. That was not a point of the original issue #463.

To be more clear, I mean I'm not opposed to options removal from attributes. It just feels related to finders only partially and probably should be considered in a separate ticket.

yratanov:
@ro0gr I've added required tests. But for `findMany` I haven't added `at` and `last`, since it doesn't seem to make any sense in this case.

`pageObjectKey` is mysterious, I found that it used in error messages, so added a test for this case.

ro0gr:
> for findMany I haven't added at and last, since it doesn't seem to make any sense in this case.

great finding! I think we need to remove their mentions from the `findMany` docs and maybe even ensure they don't have effect on a find result by picking whitelisted options and pass only them to the `getExecutionContext(pageObjectNode).find(..`. 

I think whitelisting part is not super critical, but it can be a good opportunity to have a bit cleaner API for new finders.

yratanov:
@ro0gr added for both `findOne`, `findMany`. Instead of picking only whitelisted options I throw error if something not from this list is passed. What do you think?

ro0gr:
> instead of picking only whitelisted options I throw error

I think it can be a bit ubtrosive. Typically JS is quite forgiving, and if you pass an unknow option to some object argument, an unknown option would be just ignored by a function. Throwing is not expected to me in this case. I think I'd just explicitly pick and pass known options down to the search function.

addon-test-support/properties/is-visible.js
```diff
@@ -112,15 +113,14 @@ export function isVisible(selector, userOptions = {}) {
...
+      guardMultiple(elements, selector, options.multiple);
 
       if (elements.length === 0) {
         return false;
       }
 
-      return every(elements, function(element) {
-        return element.is(':visible');
-      });
+      return A(elements).every((element) => window.getComputedStyle(element).display !== 'none');
```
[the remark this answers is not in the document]
yratanov: Then I think we shouldn't change these properties at all and leave usage of `findElementWithAssert` and `findElement`. What do you think?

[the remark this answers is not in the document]
ro0gr: I think the new finders should cover all the necessary cases needed to implement custom attrs and actions.

In this specific case, I'd keep searching with new finders, but use jquery for the check, like:

```js
  $(element).is(':visible');
```

It isn't super different from what we had before, but I still find it valuable to switch to new finders, and make dependencies on JQuery specific features more explicit(isolated) for each specific case.

ro0gr:
> I think it can be a bit ubtrosive.

@yratanov just re-read my comment 😨  That was a typo, sorry! I meant **o**btrusive for sure.

guides/extend.md
```diff
@@ -8,8 +8,76 @@ title: Extend
...
+[addon/-private/extend/find-many.js:38-44](https://github.com/san650/ember-cli-page-object/blob/c521335ffba9955a6acaf1006ed503cbb61ba72d/addon/-private/extend/find-element-with-assert.js#L38-L44 "Source code on GitHub")
+
+**Parameters**
+
+-   `pageObjectNode` **Ceibo** Node of the tree
+-   `targetSelector` **string** Specific CSS selector
+-   `options` **Object** Additional options
+    -   `options.resetScope` **boolean** Do not use inherited scope
+    -   `options.contains` **string** Filter by using :contains('foo') pseudo-class
+    -   `options.last` **boolean** Filter by using :last pseudo-class
```
ro0gr: `last` got redundant

addon-test-support/properties/value.js
```diff
@@ -93,15 +93,13 @@ export function value(selector, userOptions = {}) {
     get(key) {
       let options = assign({ pageObjectKey: key }, userOptions);
 
-      let elements = findElementWithAssert(this, selector, options);
+      const checkValue = (element) => element.hasAttribute('contenteditable') ? $(element).html() : $(element).val();
```
ro0gr: according to CI, there is a missing import for `$`

yratanov:
@ro0gr updated. Tests are failing on CI, but passing locally. Any advice? :)

ro0gr:
@yratanov Sorry me for being off for a long time. 

I just had a chance to test out your branch. It fails in `is-visible`, `is-hidden` and `is-present`. They contain a code path like:

```js
  let elements = findMany(this, selector, options);
  guardMultiple(elements, selector, options.multiple);
```

At the moment with the new `findMany` you can't use `at` and `last` anymore. So, if you defined an attribute using `at`, and there are multiple matching elements, then the `guardMultiple` would fail.

I think, even when we deprecate and remove `multiple` option from the attribute options, we would still not able to use `findOne` for this scenario, cause it's supposed to throw in case of no element was found. 

Seems like removal of `at` and `last` from the `findMany` was a wrong idea, and we should restore these for `findMany`. So, in case of such options being passed, `findMany` would return not more than 1 element in a result list. 

What do you think?

yratanov:
@ro0gr tests are green now

ro0gr:
Just realized it'd be 🔥 , if you can also add types for the new finders, alongside the old finders https://github.com/san650/ember-cli-page-object/blob/f3036572f6584dae1589134608db2dbd911c5dfd/types/index.d.ts#L63-L64, and tests into https://github.com/san650/ember-cli-page-object/blob/master/types/extend-test.ts

Please let me know if you need help with it!

Anyways, it's not blocking, I can add it as a follow up PR.

yratanov:
@ro0gr done 👌

ro0gr:
@yratanov thank you for your effort!

Issue #392, opened 2018-03-16
Improve assertions


ro0gr:
I [think](https://github.com/ember-learn/ember-cli-addon-docs/pull/112#issuecomment-373855353) there is a lack of assertions expressiveness using page objects:
```js
// .ok() results with a very poor assertion message which is context-less.
// expected: true
// actual: false
assert.ok(page.title.isVisible, 'DocsHero');

// .equal() also leads to a non-contextual message
// expected: 'DocsHero'
// actual: 'NotAHero'
assert.equal(page.title.text, 'DocsHero');
```
 I think we can get an inspiration from `qunit-dom`. It has very similar API:
 - isVisible
 - hasClass
 - hasText(contains in the page-object world)
 - lots...

What if we have own qunit extension like:

```js
// expected: page title is visible
// actual: page title is not visible
assert.PO(page.title).isVisible(); // use of PO just for visibility

// expected: page title has text "DocsHero"
// actual: page title has text "NotAHero"
assert.PO(page.title).hasText('DocsHero');

assert.PO(page.title).hasAttribute('aria-role', 'button');
```
This way an assertion can get all the necessary information to provide the best message possible

### Better integration with qunit-dom

There is also an option to connect `page-object` with `qunit-dom`. But page object doesn't provide a way to get a DOM `element`. If we even expose an `element` then assertion would become too verbose:
```js
  assert.dom(page.title.element).hasText('DocsHero');
```
Also there will not be an ability to support custom(defined by user `page.isOnTop`) page object props.

### ember-qunit-nice-errors

We can recommend to use ["ember-qunit-nice-errors"](https://github.com/wyeworks/ember-qunit-nice-errors) for better error messages. AFAIK it instruments qunit assertion lines in the build time to provide an assertion source code as an assertion message.
So for the following test:
```
test('it works', function(assert) {
  assert.equal(page.title.text, 'DocHero');
});
```
you would get the following assetion message:
```js
  assert.equal(page.title.text, 'DocHero');
  expected: 'DocHero'
  actual: 'NotAHero'
```
I feel like we still can improve readability of the tests code and assertion messages in runtime.

san650:
@ro0gr have you tried [ember-qunit-nice-errors](https://github.com/wyeworks/ember-qunit-nice-errors) ? This other addon was my attempt to solve the same problem. It's qunit specific though.

ro0gr:
[...]

PR #467, opened 2019-11-06
add a way to get native element

NullVoxPopuli:
This feels weird. Is there way a better way to get the native element?

The main thing I'm looking for is a way to integrate qunit-dom with page objects.
Right now, it feels like we have to either use ec-page-object or qunit-dom.... which makes me sad :(

### DIFF
--- a/addon-test-support/create.js
+++ b/addon-test-support/create.js
@@ -2,6 +2,7 @@ import Ceibo from 'ceibo';
 import { render, setContext, removeContext } from './-private/context';
 import { assign, getPageObjectDefinition, isPageObject, storePageObjectDefinition } from './-private/helpers';
 import { visitable } from './properties/visitable';
+import { element } from './properties/element';
 import dsl from './-private/dsl';
 
 //
@@ -46,9 +47,9 @@ import dsl from './-private/dsl';
 function buildObject(node, blueprintKey, blueprint, defaultBuilder) {
   let definition;
 
-  // to allow page objects to exist in definitions, we store the definition that 
+  // to allow page objects to exist in definitions, we store the definition that
   // created the page object, allowing us to substitute a page object with its
-  // definition during creation 
+  // definition during creation
   if (isPageObject(blueprint)) {
     definition = getPageObjectDefinition(blueprint);
   } else {
@@ -65,7 +66,7 @@ function buildObject(node, blueprintKey, blueprint, defaultBuilder) {
 
   // persist definition once we have an instance
   storePageObjectDefinition(instance, blueprintToStore);
-  
+
   return [ instance, blueprintToApply ];
 }
 
@@ -193,6 +194,8 @@ export function create(definitionOrUrl, definitionOrOptions, optionsOrNothing) {
       return chainedTree;
     }
   };
+
+  definition.element = element();
 
   // Build the primary tree
   let builder = {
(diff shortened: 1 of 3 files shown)


ro0gr:
> Is there way a better way to get the native element?

yes, I don't know if it's better, but it exists:

```js
findElement(pageObjectInstance).get(0);
findElementWithAssert(pageObjectInstance).get(0);
```

also, with https://github.com/san650/ember-cli-page-object/issues/463 it should be possible to avoid `.get(0)`:

```js
findOne(pageObjectInstance);  
```

However, these methods can be imported from a slightly unusual location:

```js
import { findElement } from 'ember-cli-page-object/extend'; // <= extend
```

which kinda hints, that these are special methods that are probably not supposed to be used in tests directly(typically).

>  it feels like we have to either use ec-page-object or qunit-dom.... 

That's true, and that's not obviously bad to me. I think these are just 2 libraries with 2 different approaches to the problem.

"qunit-dom" does an excellent job for those who use CSS selectors intensively in their test suites. 

"ember-cli-page-object" in it's turn helps to abstract CSS selectors out of your tests. I believe, if we expose an `element` as a public API in page objects, we would open a hole to violate this basic idea, so, it would be easy to do smth like:

```js
  myPage.items[0].element.querySelector('.please + .maintain > [me]')
```

Same applies to a potential use of "ember-cli-page-object" "qunit-dom":

```js
  // here we start encoding DOM knowledge in the test
  assert.dom(myPage.element).hasClass('active')

  // everything is fine from the page object perspective
  assert.equal(myPage.isActive)
```

----

FWIW, I believe while it's fine to not integrate with "qunit-dom", we still have something to learn from it. A strong benefit of "qunit-dom" is a great assertions feedback. The desire to achieve the same level of feedback in page objects has led me to https://github.com/san650/ember-cli-page-object/issues/392 and a POC of [ember-qunit-page-object](https://github.com/ro0gr/ember-qunit-page-object/blob/master/tests/smoke-test.js#L54) assertions(I hope to revive my efforts there at some point), which builds qunit assertions based on your page object shape. Also there is a [ember-qunit-nice-errors](https://github.com/wyeworks/ember-qunit-nice-errors) which improves assertion messages at build time.

NullVoxPopuli:
Halfway through reading your reply and and examples, and I was like, "you make good points, maybe the integration should be on the qunit-dom side instead"

And then you nearly said that same thing!
So, that's exciting. I wonder if we can integrate directly into qunit Dom instead of having another qunit integration library? Idk. The assertions in qunit-dom just need to understand / translate the page object apis. Or do the .get(0) thing.

ro0gr:
I'm not sure how "qunit-dom" could translate page object API, cause "qunit-dom" has a determenistic API surface(`hasClass`, `hasTag`, `includesText`), while with page objects user defines component testing APIs on their own, like `isActive`, `canDoSomething`, ...

NullVoxPopuli:
I guess, maybe we start with 

`assert.dom(po).regularAssertion()`,
And ignore the properties on the page object.

This then allows the selectors and structure to be abstracted, while being able to assert content, or whatever.

Then, maybe we make the properties passable, like:

```
import { isVisible, text } from 'ember-cli-page-object';
// ...
assert.dom(po).have(isVisible);
assert.dom(po).have(text, 'expected');
```

Idk

PR #471, opened 2019-11-10
add options to types

NullVoxPopuli:
When I deleted my custom types to use the new builtin types, I got some errors, so.. here we are :)

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -36,7 +36,7 @@ declare module 'ember-cli-page-object' {
   function triggerable(event: string, scope?: string, eventOptions?: TriggerOptions, options?: FindOptions): MethodDescriptor<<T>(this: T) => T>;
   function focusable(scope?: string, options?: FindOptions): MethodDescriptor<<T>(this: T) => T>;
   function blurrable(scope?: string, options?: FindOptions): MethodDescriptor<<T>(this: T) => T>;
-  function visitable(path: string, options?: FindOptions): MethodDescriptor<<T>(this: T, dynamicSegmentsAndQueryParams?: {}) => T>;
+  function visitable(path: string): MethodDescriptor<<T>(this: T, dynamicSegmentsAndQueryParams?: {}) => T>;
 
   interface Collection<T> {
     (scope: string, definition?: T): Array<Component<T>>;
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -20,11 +20,11 @@ declare module 'ember-cli-page-object' {
   function isPresent(scope?: string, options?: FindOptions): GetterDescriptor<boolean>;
   function text(scope?: string, options?: FindOptions): GetterDescriptor<string>;
   function value(scope?: string, options?: FindOptions): GetterDescriptor<string>;
-  function property(name: string, options?: FindOptions): GetterDescriptor<any>;
+  function property(name: string): GetterDescriptor<any>;
   function property(scope: string, name: string, options?: FindOptions): GetterDescriptor<any>;
-  function hasClass(className: string, options?: FindOptions): GetterDescriptor<boolean>;
+  function hasClass(className: string): GetterDescriptor<boolean>;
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -14,29 +14,29 @@ declare module 'ember-cli-page-object' {
...
   function clickOnText(scope?: string, userOptions?: FindOptions): MethodDescriptor<<T>(this: T, text: string) => T>;
   function fillable(scope?: string, userOptions?: FindOptions): MethodDescriptor<<T>(this: T, clueOrContent: string, content?: string) => T>;
-  function triggerable(event: string, scope?: string, eventOptions?: TriggerOptions): MethodDescriptor<<T>(this: T) => T>;
-  function focusable(scope?: string): MethodDescriptor<<T>(this: T) => T>;
-  function blurrable(scope?: string): MethodDescriptor<<T>(this: T) => T>;
-  function visitable(path: string): MethodDescriptor<<T>(this: T, dynamicSegmentsAndQueryParams?: {}) => T>;
+  function triggerable(event: string, scope?: string, eventOptions?: TriggerOptions, options?: FindOptions): MethodDescriptor<<T>(this: T) => T>;
+  function focusable(scope?: string, options?: FindOptions): MethodDescriptor<<T>(this: T) => T>;
+  function blurrable(scope?: string, options?: FindOptions): MethodDescriptor<<T>(this: T) => T>;
+  function visitable(path: string, options?: FindOptions): MethodDescriptor<<T>(this: T, dynamicSegmentsAndQueryParams?: {}) => T>;
```
ro0gr: I think visitable doesn't support `FindOptions`

NullVoxPopuli: whoops

types/index.d.ts
```diff
@@ -14,28 +14,28 @@ declare module 'ember-cli-page-object' {
...
-  function notHasClass(scope: string, className: string): GetterDescriptor<boolean>;
-  function contains(scope?: string): (text: string) => GetterDescriptor<boolean>;
-  function count(scope?: string): () => GetterDescriptor<boolean>;
+  function attribute(attributeName: string, scope?: string, options?: FindOptions): GetterDescriptor<string>;
+  function isVisible(scope?: string, options?: FindOptions): GetterDescriptor<boolean>;
+  function isHidden(scope?: string, options?: FindOptions): GetterDescriptor<boolean>;
+  function isPresent(scope?: string, options?: FindOptions): GetterDescriptor<boolean>;
+  function text(scope?: string, options?: FindOptions): GetterDescriptor<string>;
+  function value(scope?: string, options?: FindOptions): GetterDescriptor<string>;
+  function property(name: string, options?: FindOptions): GetterDescriptor<any>;
```
ro0gr: looking at the current implementation https://github.com/san650/ember-cli-page-object/blob/master/addon-test-support/properties/property.js#L69

I seems like it would not workm cause with such signatire it would call smth like:

```
findElementWithAssert(this, findOptions, {});
```
which is not expected.

I think this line should stay unchanged

types/index.d.ts
```diff
@@ -14,28 +14,28 @@ declare module 'ember-cli-page-object' {
...
+  function property(scope: string, name: string, options?: FindOptions): GetterDescriptor<any>;
+  function hasClass(className: string, options?: FindOptions): GetterDescriptor<boolean>;
```
ro0gr: same as for `property`. I think we should leave a single arg signature as is.

types/index.d.ts
```diff
@@ -14,28 +14,28 @@ declare module 'ember-cli-page-object' {
...
+  function hasClass(scope: string, className: string, options?: FindOptions): GetterDescriptor<boolean>;
+  function notHasClass(className: string, options?: FindOptions): GetterDescriptor<boolean>;
```
ro0gr: +1

NullVoxPopuli: yeah, must have gotten carried away with the copy-paste

ro0gr:
Awesome! Thank you!

ro0gr:
just published as 1.16.1

PR #478, opened 2020-01-21
Convert test suite to TS

ro0gr:
This would allow to truly test our public APIs, w/o relying on synthetic dtslint tests.
An intention is to only convert tests which don't use deprecated(or almost deprecated) APIs. Even this way quite few typings issues were detected and some of them fixed in this PR.

Unfortunately I had to pin to ec-ts@1, cause on a later versions ec-babel@7 or higher is required, and it requires node@6 or higher. And that would be a breaking change, cause currently we have node@4.5 listed in our engines. So upgrade of ec-ts should be done with the next major version.

todo: 
 - [x] finish migration and make it green
 - [x] make sure if we don't publish any of new TS stuff


tsconfig.json
```diff
@@ -0,0 +1,56 @@
...
+    "noImplicitAny": true,
+    "noImplicitThis": true,
+    "alwaysStrict": true,
+    "strictNullChecks": true,
+    "strictPropertyInitialization": true,
+    "noFallthroughCasesInSwitch": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noImplicitReturns": true,
+    "noEmitOnError": true,
```
ro0gr: This fails the build in case of any TS build error occurs for both development and CI. 
It may be not convenient for everyone, especially for people who are new with TS, or those who just want to quickly experiment and not to fight with types. 

So, if anyone knows a way to only enable such behavior on Travis, hints/suggestions are welcome!

PR #492, opened 2020-03-29
v2: Remove find element

ro0gr:
and remove `mutliple` find option as not used anymore


ChristianGreinke:
@ro0gr 
Just saying we have been using {multiple:true} all over our tests. So maybe it's only us but still "and remove mutliple find option as not used anymore" would be kinda wrong if there is at least one case. Further removing public api implementation without any strong benefit is not a great way to handle improvements. In my opinion at least. Anyway appreciate your investment into improving this lib.

PR #493, opened 2020-03-29
Prepare for adapters

ro0gr:
This PR is focused on the:

> A noticeable difference to ExecutionContexts is that Adapter is no longer responsible for the Element search logic, which requires an action to find the element before passing it to adapter.

of the https://github.com/san650/ember-cli-page-object/issues/479.

Also in this PR we drop `findElementWithAssert` code as not needed anymore.


ro0gr:
close in favour of #509

PR #510, opened 2020-06-20
extract alias utils from "helpers"

ro0gr:
a part of #512

"helpers.js" is currently polluted with some of single-use utils. Moving such utils to a relevant context shuold imporve code clarity a bit. 

Also removed utils tests, since the use cases are covered via alias tests


addon-test-support/macros/alias.js
```diff
@@ -106,3 +103,59 @@ export function alias(pathToProp, options = {}) {
...
+  if (pathSegments.length === 1) {
+    const value = get(object, pathToProp);
+    return typeof value === 'function' ? value.bind(object) : value;
+  }
+
+  const pathToPropOwner = pathSegments.slice(0, -1).join('.');
+  const propOwner = get(object, pathToPropOwner);
+
+  if (propOwner === null || propOwner === undefined) {
+    return undefined;
```
ro0gr, line 154: lost coverage here

PR #511, opened 2020-06-20
Adapters

ro0gr:
Implements https://github.com/san650/ember-cli-page-object/issues/479

With this PR we convert private ExecutionContexts to public Adapters.

It's required now to `setAdapter` before start using page objects in tests:

```js
// tests/test-helper.js

import DefaultAdapter from 'ember-cli-page-object/adapters/rfc268';
import { setAdapter } from 'ember-cli-page-object/adapters';

setAdapter(new DefaultAdapter());
```

in case you you need multiple adapters being set for different modules,
make sure to set a teardown it before each test:

```js
  hooks.beforeEach(function() {
    setAdapter(new MyAdapter());
  });

  hooks.afterEach(function() {
    setAdapter(new DefaultAdapter());
  });
```

if no adapter is set, an error throws.

This PR does also remove few of the old and deprecated ModuleFor-only related APIs, like:

     - `getContext(`, `setContext(`,`removeContext(` - can be  replaced with `setAdapter(`
     - `render(`
     - `page.context`
     - `useNativeEvents(` - can be  replaced with `setAdapter(`
     - `andThen` behavior support. Now you have to explicitly wait for the previous action completeness.

And some APIs that just don't make sense anymore:

     - `registerExecutionContext`


ro0gr:
There are some work-arounds for old test scenarios. Though in general I'd like to get rid of legacy stuff like `moduleFor*` adapters, and old Ember version in scope of v2. However, I'd prefer to make sure it's possible to support via user land adapters, before removal from the code base.

tests/acceptance/rfc268-test.js
```diff
@@ -0,0 +1,34 @@
+import { module, test } from 'qunit';
+import require from 'require';
+import Rfc268Adapter from 'ember-cli-page-object/adapters/rfc268';
+import { setAdapter } from 'ember-cli-page-object/adapters';
+import { create, value, visitable } from 'ember-cli-page-object';
+
+if (require.has('@ember/test-helpers')) {
+  // intentionally not using our local extension in order to make
+  // sure, RFC268 works by default, w/o Adapter being set.
```
ro0gr, line 9: copy-pasted comment not applicable

PR #490, opened 2020-03-29
bump version jquery

mum-never-proud:
Resolves #456

PR #513, opened 2020-06-30
unpin jquery

kellyselden:
I think this was a mistake in https://github.com/san650/ember-cli-page-object/pull/490. This prevents us from getting the latest jquery without using yarn resolutions or forking.


ro0gr:
Thanks! I'll publish it today evening.

ro0gr:
published as 1.17.3

PR #515, opened 2020-10-09
add :is to types

NullVoxPopuli:


types/index.d.ts
```diff
@@ -28,6 +28,7 @@ declare module 'ember-cli-page-object' {
   function notHasClass(scope: string, className: string, options?: FindOptions): GetterDescriptor<boolean>;
   function contains(scope?: string, options?: FindOptions): (text: string) => GetterDescriptor<boolean>;
   function count(scope?: string, options?: FindOptions): () => GetterDescriptor<boolean>;
+  function is(stateSelector: string, options?: FindOptions): GetterDescriptor<boolean>;
```
ro0gr, line 31: Can we maybe add `@deprecated` JSDoc annotation?

ro0gr (commented):
Thanks for this! It looks good to me, however, `is` is deprecated, and it's going to be removed in v2. I think we should highlight to the user. 

Also I'd appreciate ["tests/unit/-private/properties/is-test.js"](https://github.com/san650/ember-cli-page-object/blob/master/tests/unit/-private/properties/is-test.js) to be converted to TS, in order to make sure, that the new typings are correct.

ro0gr:
the `is(` is not a part of the master anymore

PR #516, opened 2020-10-23
Compatibility with ember-qunit@5.0.0-beta

rohitpaulk:


ro0gr:
Thanks! it makes perfect sense to me to cover new version of the ember-qunit.

Unfortunately, seems like currently CI is broken for some reason. I'll take a look and try to fix it this week.

ro0gr:
just retargeted it to the v2-beta, but Travis seems to be weird last months, and I can't see aby builds re-triggered here https://travis-ci.org/github/san650/ember-cli-page-object/pull_requests. I'd be glad to test it locally, but first I have some questions..

config/ember-try.js
```diff
@@ -154,6 +154,23 @@ module.exports = function() {
             }
           }
         },
+        {
+          name: 'with-ember-qunit@5.0.0-beta',
+          bower: {
+            dependencies: {
+              'ember': 'components/ember#release'
```
ro0gr, line 161: why do we use bower?

ro0gr: can't we use the same treatment as https://github.com/san650/ember-cli-page-object/blob/484e3fccbee26d501e3dfecfdfa110094ff6c131/config/ember-try.js#L78-L83?

rohitpaulk:
@ro0gr I'm guessing the reason Travis isn't running is because of the `branch: master` filter [here](https://github.com/san650/ember-cli-page-object/blob/4828fc116baf8348fbfca8207ba2e3b8e4ad55a7/.travis.yml#L19-L21). Do you think it'd be okay to include `v2-beta` to that list? 

I'll look into the bower stuff - iirc I just copied what was in a configuration above, it's possible that some of it is dated.

ro0gr:
ah... you a right. I've just retargeted it via GH interface, but seems your branch is still build on top of master. In v2-beta we have the `travis` config configured properly https://github.com/san650/ember-cli-page-object/blob/v2-beta/.travis.yml#L21-L22

Can you please rebase your branch on top of beta-v2

ro0gr:
#530 has been merged. Thanks!

PR #517, opened 2020-10-27
Use node@10 on travis, to appease qunit

ro0gr:
unsupported node versions were dropped by QUnit recently, see: https://github.com/qunitjs/qunit/pull/1464


ro0gr:
close in favour of #520

PR #520, opened 2020-11-14
Set Node.js version to 10 in order to fix TravisCI build

gnclmorais:
I’ve tried to follow [this guide](https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md#node-support) to set different Node versions according to the Ember CLI version but that didn’t work out so well ([here’s an example build](https://travis-ci.org/github/gnclmorais/ember-cli-page-object/builds/743370770)).

Setting the Node version to 10 [fixed the building process](https://travis-ci.org/github/gnclmorais/ember-cli-page-object/builds/743575169) (commit de132c2 allowed me to build branches that are not `master`).

This pull request’s commits can be squashed to only two: one sets the Node version to 10 and the other one (e8d872a) addresses a few issues with TravisCI configuration. Let me know what you think! 😃 

https://travis-ci.org/github/gnclmorais/ember-cli-page-object/builds/743575169

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,10 @@ env:
     # See https://git.io/vdao3 for details.
     - JOBS=1
 
+branches:
+  only:
+    - master
+
 jobs:
   fast_finish: true
   allow_failures:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,6 @@ env:
     # See https://git.io/vdao3 for details.
     - JOBS=1
 
-branches:
-  only:
-    - master
-
 jobs:
   fast_finish: true
   allow_failures:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-  - "8"
+  - 12
 
 addons:
   chrome: stable
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - 12
+  - lts/*
 
 addons:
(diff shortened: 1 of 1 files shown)


ro0gr:
@gnclmorais Thank you a lot for your effort! Great job!

Yes, I've also wanted to preserve testing against the older Node.js versions, at least for some of scenarios, but seems like there are some transitive dependencies which break compatibility. For instance node@6 [scenario](https://travis-ci.org/github/gnclmorais/ember-cli-page-object/jobs/743370773#L774) you've provided:

> error detect-indent@6.0.0: The engine "node" is incompatible with this module. Expected version ">=8".

And `npm list detect-indent` shows that it comes from the ember-cli:

> ├-┬ ember-cli@3.5.1 
> │ └─┬ sort-package-json@1.40.0 
> │   └── detect-indent@6.0.0 

So seems like the best we can do is to just go ahead and bump Node.js version. I hope we would not have any changes in our build code, and blueprints in scope of v1, so we'd theoretically keep working against older node.

> This pull request’s commits can be squashed to only two

I'm not sure if GH allows to squash into 2 commits. Please feel free to squash it as you wish. Otherwise, I'm going to merge it squashed into one commit next few days.

gnclmorais:
Thanks for the explanation @ro0gr, that makes total sense. I have nothing against squashing this into a single commit, just wanted to leave a different option open in case people responsible for this project preferred it. Go ahead and merge as one commit whenever you want. :) Thanks!

ro0gr:
Thanks!

Issue #518, opened 2020-10-29
Clicking on text with comma produces a deprecation warning


veelenga:
Example:

```js
// <fieldset>
//  <button>Lorem, Ipsum</button>
// </fieldset>

import { create, clickOnText } from 'ember-cli-page-object';

const page = create({
  clickOnButton: clickOnText('button')
});

page.clickOnButton('Lorem, Ipsum'); // Usage of comma separated selectors is deprecated in ember-cli-page-object
```

ro0gr:
Nice catch! This is definitely unexpected. I think a failing test can help to move it forward.

PR #494, opened 2020-03-29
Disallow comma-separated selectors

ro0gr:
needs
  - [ ] test for findOne and findMany

PR #521, opened 2020-11-21
Do not warn when selecting a text with comma

veelenga:
When a text with comma is selected, the ember-cli-page-object produces a warning. Example:

```javascript
// <fieldset>
//  <button>Lorem, Ipsum</button>
// </fieldset>

import { create, clickOnText } from 'ember-cli-page-object';

const page = create({
  clickOnButton: clickOnText('button')
});

page.clickOnButton('Lorem, Ipsum'); // Usage of comma separated selectors is deprecated in ember-cli-page-object

```

Fixes https://github.com/san650/ember-cli-page-object/issues/518
Conflicts with https://github.com/san650/ember-cli-page-object/pull/494

### DIFF
--- a/addon-test-support/-private/helpers.js
+++ b/addon-test-support/-private/helpers.js
@@ -27,6 +27,15 @@ class Selector {
       scope = this.calculateScope(this.targetNode, this.targetScope);
     }
 
+    deprecate(
+      'Usage of comma separated selectors is deprecated in ember-cli-page-object',
+      `${scope} ${this.targetSelector}`.indexOf(',') === -1, {
+        id: 'ember-cli-page-object.comma-separated-selectors',
+        until: "2.0.0",
+        url: 'https://ember-cli-page-object.js.org/docs/v1.16.x/deprecations/#comma-separated-selectors',
+      }
+    );
+
     filters = this.calculateFilters(this.targetFilters);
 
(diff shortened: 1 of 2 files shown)


addon-test-support/-private/helpers.js
```diff
@@ -38,7 +38,7 @@ class Selector {
     }
 
     deprecate(
-      'Usage of comma separated selectors is deprecated in ember-cli-page-object', selector.indexOf(',') === -1, {
+      'Usage of comma separated selectors is deprecated in ember-cli-page-object', !this.isCommaSeparated(selector), {
```
ro0gr, line 41: I think the problem here is that we just do the check too late, when the `targetFilters` are applied already.
Can we move the check somewhere before it happened https://github.com/san650/ember-cli-page-object/pull/521/files#diff-7f9866fcc737dac901e9ba9b15adae0045477e229889aaf75ec3f343550d45f0R32, in order to avoid working around the filters?

ro0gr (commented):
Thank you!

I'd like to avoid an extra method, by doing the check a bit earlier.

ro0gr (approved):
Excellent! Thanks!

ro0gr:
just published your fix as 1.17.4

PR #528, opened 2021-01-07
Upgrade to ember@3.23

ro0gr:
...and remove legacy adapters due to amount of issues with support for ember from bower and other issues aboiut old versions of ember with the new ember-cli. Seems like there opens a window for the future tests refactorings.

one of the rationale for the PR is a desire to replace `assign` util with the spread operator, as the next step.


ro0gr:
replaced by https://github.com/san650/ember-cli-page-object/pull/560

PR #534, opened 2021-02-05
Allow undefined urls to be undeprecated

mixonic:
In https://github.com/san650/ember-cli-page-object/commit/7125ba813dc2ecac361828575ca2781cefa364c7 a deprecation was added for `create('')`, however it also logs an error when used with `create()`. At https://github.com/san650/ember-cli-page-object/commit/7125ba813dc2ecac361828575ca2781cefa364c7#diff-7e9135d873a4889a025f4ed419cf7d3025a1fc38f7ec08bb1c7b3800c55a3483L200 a value of `false` may be set to `url`, which means `create()` should *always* be firing a deprecation.

I see there is a test in that same patch, I haven't confirmed it is running or not running though https://github.com/san650/ember-cli-page-object/commit/7125ba813dc2ecac361828575ca2781cefa364c7#diff-3963c3a74ca7d2852220eb87172db93f91df773a766ea56f2d3c92a42df646c4R12

### DIFF
--- a/addon-test-support/create.js
+++ b/addon-test-support/create.js
@@ -197,7 +197,6 @@ export function create(definitionOrUrl, definitionOrOptions, optionsOrNothing) {
     definition = definitionOrOptions || {};
     options = optionsOrNothing || {};
   } else {
-    url = false;
     definition = definitionOrUrl || {};
     options = definitionOrOptions || {};
   }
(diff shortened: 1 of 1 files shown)


ro0gr:
> however it also logs an error when used with `create()`

I've just checked the test you've mentioned, and I believe it works fine(w/o a deprecation) for `create()`, can you help me understand your scenario better please?

mixonic:
I'm closing this, if I have a better reproduction I'll re-open.

PR #540, opened 2021-02-10
Fix `Ember.emberDeprecate` is not a function

alexlafroscia:
The internal `deprecate` wrapper imports Ember's deprecation utility from a different location than the documented one.

When running this addon against the latest version of Ember, a runtime error occurs where Ember's deprecation function is not defined.

`deprecate` documentation: https://api.emberjs.com/ember/3.24/functions/@ember%2Fdebug/deprecate

### DIFF
--- a/addon-test-support/-private/deprecate.js
+++ b/addon-test-support/-private/deprecate.js
@@ -1,4 +1,4 @@
-import { deprecate as emberDeprecate } from '@ember/application/deprecations';
+import { deprecate as emberDeprecate } from '@ember/debug';
 
 const NAMESPACE = 'ember-cli-page-object';
 
--- a/addon-test-support/-private/deprecate.js
+++ b/addon-test-support/-private/deprecate.js
@@ -1,11 +1,11 @@
-import { deprecate as emberDeprecate } from '@ember/debug';
+import { deprecate } from '@ember/debug';
 
 const NAMESPACE = 'ember-cli-page-object';
 
-export default function deprecate(name, message, since, until) {
+export default function deprecateWrapper(name, message, since, until) {
   const [major, minor] = since.split('.');
 
-  emberDeprecate(message, false, {
+  deprecate(message, false, {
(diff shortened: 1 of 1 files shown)


alexlafroscia:
For some reason, when the `deprecate` import is renamed through the import, the Babel plugin breaks. It converts the function into `Ember.emberDeprecate` rather than `Ember.deprecate`.

At first I thought that this was all caused by importing from the wrong module (my first commit) but that didn't actually fix the problem.

I then figured this would be a bug in `ember-cli-babel` since the version used here is really old. However, I updated the dependency and that didn't fix the issue. 

Just avoiding the alias while importing _does_ fix the problem so... I guess we'll do that? 🤷‍♂️

alexlafroscia:
Interestingly, it seems like the `@ember/debug` modules are handled as a special-case in `ember-cli-babel`...

https://github.com/babel/ember-cli-babel/blob/76876904812fee227fff2af1e29aca73cf6f1c76/node-tests/addon-test.js#L113-L136

I wonder if that is related to the fact that the alias on importing seems broken.

alexlafroscia:
Oh hey, here's the actual bug

https://github.com/ember-cli/babel-plugin-debug-macros/issues/48

ro0gr:
Hey, thanks for the fix and info!

I'm going to merge and publish it, once CI passes.

ro0gr:
Recently I had to do a similar change to the v2-beta branch https://github.com/san650/ember-cli-page-object/pull/537/commits/2d7ea1b22f3ac910381c2f0b009fc6d9db293691, but I didn't realized we do use invalid import path. Will fix it there as well.

ro0gr:
alright, travis build has failed for ember-beta scenario due to a monstrous amount of deprecations, and maximum log length 😋 
seems I'd need to cleanup some of ember deprecations to make Travis a bit more happy. But not today.

ro0gr:
published as 1.17.7. Thanks!

Issue #514, opened 2020-09-11
content with display: none is still captured (does not match innerText)


sukima:
The `PageObject.text` helper returns hidden text. While — by comparison — `innerText` does not.

## Given

```hbs
<span id="example">
  <span>foo</span>
  <span style="display: none">bar</span>
  <span>baz</span>
</span>
```

```js
const page = PageObject.create({
  example: PageObject.text('#example')
});
```

## Expected

```js
find('#example').innerText.trim() // => "foo baz"
page.example                      // => "foo baz"
```

## Actual

```js
find('#example').innerText.trim() // => "foo baz"
page.example                      // => "foo bar baz"
```

sukima:
This is how I am getting around this currently.

```js
const page = create({
  example: getter(function() {
    let [element] = findOne(this, '#example');
    return element.innerText.trim();
  })
});
```

sukima:
And just in case someone wonders why this is an issue here is an example use case:

### With this CSS

```css
.auto-hiding-delimeter:first-child,
.auto-hiding-delimeter:last-child,
.auto-hiding-delimeter + .auto-hiding-delimeter {
  display: none;
}
```

### You can do things like these

```hbs
{{#each @things as |thing|}}
  {{#if thing}}<span class="thing">{{thing}}</span>{{/if}}
  <span class="auto-hiding-delimiter">[seperator]</span>
{{/each}}
```

```hbs
{{#if this.boolFlagOne}}<span>1</span>{{/if}}
<span class="auto-hiding-delimiter">|</span>
{{#if this.boolFlagTwo}}<span>2</span>{{/if}}
<span class="auto-hiding-delimiter">|</span>
{{#if this.boolFlagThree}}<span>3</span>{{/if}}
<span class="auto-hiding-delimiter">|</span>
{{#if this.boolFlagFour}}<span>4</span>{{/if}}
<span class="auto-hiding-delimiter">|</span>
{{#if this.boolFlagFive}}<span>5</span>{{/if}}
```

ro0gr:
Thanks for reporting! 

This is surprising to me. I didn't know that jquery uses `textContent` for the `text(`  https://github.com/jquery/jquery/blob/e35fb62db4fb46f031056bb53e393982c03972a1/src/core.js#L276.

I think replacing our current usage of the `$.text(` with the `innerText` would be a breaking change, even though it can be treated as a **fixing** breaking change.. So I tend to avoid it in v1.

[...]

PR #547, opened 2021-10-03
fix: Revert native contains property

Alonski:
The change done in this PR: #466 caused a regression.

$.text() can find the text in `visibility: hidden` or `display: none` elements.

element.innerText does not return the text. 

Fixes: #514

### DIFF
--- a/addon-test-support/properties/contains.js
+++ b/addon-test-support/properties/contains.js
@@ -1,6 +1,5 @@
-import { assign } from '../-private/helpers';
-import { findMany, findOne } from '../extend';
-import { A } from '@ember/array';
+import { assign, every } from '../-private/helpers';
+import { findElementWithAssert } from '../extend';
 
 /**
  * Returns a boolean representing whether an element or a set of elements contains the specified text.
@@ -69,7 +68,6 @@ import { A } from '@ember/array';
  *
  * const page = create({
  *   scope: '.scope',
-
  *   spanContains: contains('span')
  * });
  *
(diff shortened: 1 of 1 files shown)


addon-test-support/properties/contains.js
```diff
@@ -1,6 +1,5 @@
-import { assign } from '../-private/helpers';
-import { findMany, findOne } from '../extend';
-import { A } from '@ember/array';
+import { assign, every } from '../-private/helpers';
+import { findElementWithAssert } from '../extend';
```
ro0gr, line 2: internaly we tend to avoid using `findElementWithAssert(` in favour of `findMany` or `findOne`. I'd appreciate if we can avoid `findElementWithAssert` here as well.

ro0gr (commented):
Thanks for moving it forward! Mind adding a test case for this regression? Some unit test [here](https://github.com/Alonski/ember-cli-page-object/blob/master/tests/unit/-private/properties/contains-test.ts) should be fine.

ro0gr:
So turns out it doesn't really fix the https://github.com/san650/ember-cli-page-object/issues/514#issuecomment-942697666, which has never been a regression. But it fixes a real regression of the `contains(`. Going to squash the commits excluding the `Fixes: #514` part.

Alonski:
@ro0gr @san650 I believe the Ember community is moving towards Github Actions vs Travis in the future. So migrating to Actions isn't a bad idea 💡

Thanks for merging and releasing!

Alonski:
@ro0gr It seems that now I added a regression in this PR 🙃 

I added a dependency to the `every` helper function on jQuery which breaks in apps that don't have jQuery integration set to true.

I need to think of a way to fix this as we also can't use `element.text()` now.

Alonski:
@ro0gr Created a reproduction in StackBlitz:
https://stackblitz.com/edit/github-oujqsi?devtoolsheight=33&file=package.json

In console:
```sh
npm i
npm run test:ember --server
```

image

Alonski:
@ro0gr Can you please also add the `hacktoberfest-accepted` label to this PR as well :)

PR #555, opened 2021-12-16
Update the names of the find methods for collection

DingoEatingFuzz:
The docs said find and findBy but the actual names are findOne and findOneBy.


ro0gr:
Thanks! And sorry for the delay. Will update docs site shortly.

ro0gr:
updated http://ember-cli-page-object.js.org/docs/v1.17.x/api/collection

PR #565, opened 2022-01-26
restore RFC268 adapter as a default

ro0gr:
This behavior has been removed in the daa0ba0efceacde6cc4eb8315a684b88472216fa
due to desire to support environments other than ember.
However, this is an ember specific addon, so it should not cause
issues to ember users.

### DIFF
--- a/addon-test-support/adapter.js
+++ b/addon-test-support/adapter.js
@@ -1,23 +1,23 @@
 export default class Adapter {
   get testContainer() {
-    throw new Error('`testContainer` is not implemented for the adater');
+    throw new Error('`testContainer` is not implemented for the adapter');
   }
   visit(/* path */) {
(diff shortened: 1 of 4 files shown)


tests/test-helper.js
```diff
@@ -4,9 +4,15 @@ import * as QUnit from 'qunit';
...
 import { setup } from 'qunit-dom';
 import { start } from 'ember-qunit';
+import { setAdapter } from 'ember-cli-page-object/adapters';
+import RFC268Adapter from 'ember-cli-page-object/adapters/rfc268';
 
 setApplication(Application.create(config.APP));
 
 setup(QUnit.assert);
 
+QUnit.testStart(() => {
```
ro0gr, line 14: ✂️

Issue #571, opened 2022-04-08
Assign is being referenced from @ember/polyfills. This prevents upgrade to Ember 4


cah-brian-gantzler:
Assign is being exported here https://github.com/san650/ember-cli-page-object/blob/master/addon-test-support/-private/helpers.js#L1. Why would we be exporting assign? This is being transpiled to below. This is preventing an upgrade to ember 4.
Could we remove this import please

![image](https://user-images.githubusercontent.com/1529286/162498613-23027aa9-9cd7-4faf-943b-f07b1ba47189.png)

(Updated. Thought it was an import, turns out to be an export, which is even weirder)

cah-briangantzler:
Ok I see that this assign is imported in many other files. Could this line be replaced with?
```js
export { Object.assign as assign };
```

cah-briangantzler:
Appears this is fixed in PR https://github.com/san650/ember-cli-page-object/pull/536/files but this was merged into the V2-beta branch

Since this branch hasnt been touched since Jan 30, any way we could maybe backport this fix so consumers can upgrade to Ember 4 without waiting on the V2 version of this addon?

ro0gr:
Sorry for delay. 

Yes, I'd prefer to use the object spread approach as you referred above. Unfortunately, I doubt if it can be just cherry-picked, cause of many changes in beta happened and conflicts are possible. I think we still can do it manually though, w/o a big issue.

ro0gr:
Just released the fix as a part of https://github.com/san650/ember-cli-page-object/releases/tag/v1.17.12

PR #572, opened 2022-04-08
Patch assign from @ember/polyfills until Addon V2 is released

cah-briangantzler:
Temp patch for https://github.com/san650/ember-cli-page-object/issues/571

The V2 version fixes this in a different way

### DIFF
--- a/addon-test-support/-private/helpers.js
+++ b/addon-test-support/-private/helpers.js
@@ -1,4 +1,4 @@
-export { assign } from '@ember/polyfills';
+
 import { A } from '@ember/array';
 import { assert } from '@ember/debug';
 import { get } from '@ember/object';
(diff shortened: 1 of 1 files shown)


addon-test-support/-private/helpers.js
```diff
@@ -9,6 +9,9 @@ import { getContext as getEmberTestHelpersContext } from './compatibility';
 
 import $ from '-jquery';
 
+let assign = Object.assign;
```
ro0gr, line 12: This is technically a breaking change, since IE doesn't have `Object.assign`. I believe destructuring assignment is a better way to handle most of the `assign` use cases.

_BTW, you might want to consider v2.0.0-beta where this is solved already._

cah-briangantzler: Ahh, I always saw where Object.Assign was suggested. Using the destructure would also be a breaking change though as you would no longer export assign. Technically only meant for internal but someone could have used it. 

Thanks. WIll look at using the beta.2 Try to avoid betas in my app, but since this addon is a dev only concern, might do it.

Issue #592, opened 2022-11-02
Replace usage of ember-native-dom-helpers by @ember/test-helpers


Frank3K:
ember-cli-page-object v1.17.11 relies on [ember-native-dom-helpers](https://github.com/cibernox/ember-native-dom-helpers). The ember-native-dom-helpers package is end-of-life and relies on ember-cli-babel@6, which is a problem for upgrading to Ember 4.

A solution would be to replace the use of ember-native-dom-helpers by [@ember/test-helpers](https://github.com/emberjs/ember-test-helpers), as it is the suggested way forward from ember-native-dom-helpers: https://github.com/cibernox/ember-native-dom-helpers/issues/147.

ro0gr:
Hey @Frank3K 👋  

It's unlikely, that ember-native-dom-helpers will be removed in scope of v1. I'd suggest to give a try to the [v2.0.0-beta.4](https://github.com/san650/ember-cli-page-object/releases/tag/v2.0.0-beta.4) which doesn't have ember-native-dom-helpers as a dependency anymore.

PR #593, opened 2022-11-07
Fix jQuery detection when @ember/jquery is only an indirect dependency

Frank3K:
We took [the advice](https://github.com/san650/ember-cli-page-object/issues/592#issuecomment-1301011970) to upgrade to v2.0.0-beta. Overall it seems pretty stable already :+1:. We're currently running into one issue, which is addressed in this PR.

---

This solves the following use case:

1. App A has a dependency on addon B.
2. App A does not have a direct dependency on @ember/jquery.
3. Addon B does have a direct dependency on @ember/jquery.

Previously, it was in this case incorrectly assumed that window.jQuery would be available when running app A.

### DIFF
--- a/addon/src/-private/helpers.js
+++ b/addon/src/-private/helpers.js
@@ -1,14 +1,10 @@
 import Ceibo from '@ro0gr/ceibo';
 import deprecate from './deprecate';
-import {
-  dependencySatisfies,
-  macroCondition,
-  importSync,
-} from '@embroider/macros';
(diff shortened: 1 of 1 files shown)


ro0gr:
The change looks reasonable to me. Thanks!

ro0gr:
https://github.com/san650/ember-cli-page-object/releases/tag/v2.0.0-beta.5

Issue #613, opened 2023-07-17
Fix support for @tsconfig/ember 3.0.0 error TS7016


mayatron:
Trying to use Ember CLI Page Objects 2.0.0 with TypeScript 5.1.6 throws the following error:

```
error TS7016: Could not find a declaration file for module 'ember-cli-page-object'.
'node_modules/ember-cli-page-object/dist/index.js' implicitly has an 'any' type.
There are types at 'node_modules/ember-cli-page-object/types.d.ts', but this result
could not be resolved when respecting package.json "exports". The 'ember-cli-page-object'
library may need to update its package.json or typings.
```

This seems to be related to [this issue](https://github.com/microsoft/TypeScript/issues/52363).

Updating the page objects `package.json` locally (as suggested in the issue above) does appear to address the issue for me, but I'm not sure about any side effects or backwards-compatibility issues the change might cause.

bendemboski:
Further context -- I believe this is specifically related to using `"moduleResolution": 'bundler'`, and this patch (applied via `patch-package` or similar) fixes it for me:

```patch
diff --git a/node_modules/ember-cli-page-object/package.json b/node_modules/ember-cli-page-object/package.json
index cddb866..e92c917 100644
--- a/node_modules/ember-cli-page-object/package.json
+++ b/node_modules/ember-cli-page-object/package.json
@@ -25,7 +25,10 @@
   "license": "MIT",
   "author": "Santiago Ferreira",
   "exports": {
-    ".": "./dist/index.js",
+    ".": {
+      "import": "./dist/index.js",
+      "types": "./types.d.ts"
+    },
     "./*": "./dist/*",
     "./addon-main.js": "./addon-main.js"
   },
```

ro0gr:
Hey, thanks guys 👋

It feels reasonable for me to provide a few test scenarios to check the build against a couple of  major typescript versions. Probably via ember-try scenarios...

I'll appreciate a PR! Otherwise, I'll try to take a look on a fix this weekend.

mayatron:
When I update `package.json` as follows:

```js
"exports": {
  ".": {
    "import": "./dist/index.js",
    "types": "./types.d.ts"
  },
  "./*": "./dist/*",
  "./addon-main.js": "./addon-main.js"
},
```

... then `tsc --noEmit` passes, but `ember build` fails with the following error:

```
[...]

PR #614, opened 2023-07-26
Add support for @tsconfig/ember 3.0.0

mayatron:
- Add support for @tsconfig/ember 3.0.0
- Addresses https://github.com/san650/ember-cli-page-object/issues/613

### DIFF
--- a/test-app/package.json
+++ b/test-app/package.json
@@ -36,6 +36,7 @@
     "@embroider/test-setup": "^1.8.0",
     "@glimmer/component": "^1.1.2",
     "@glimmer/tracking": "^1.1.2",
+    "@tsconfig/ember": "^2.0.0",
     "@types/ember": "^3.1.1",
     "@types/ember-qunit": "^3.4.13",
     "@types/ember-resolver": "^5.0.10",
--- a/test-app/tests/unit/-private/properties/create-test.ts
+++ b/test-app/tests/unit/-private/properties/create-test.ts
@@ -94,7 +94,7 @@ module('create', function () {
       // @ts-expect-error violate types to check if it fails in weakly-typed envs
       create('');
       assert.true(false, 'should error');
-    } catch (e) {
+    } catch (e: any) {
       assert.strictEqual(e.message, 'Definition can not be a string');
     }
   });
--- a/test-app/tsconfig.json
+++ b/test-app/tsconfig.json
@@ -1,47 +1,18 @@
 {
+  "extends": "@tsconfig/ember/tsconfig.json",
   "compilerOptions": {
-    "target": "es2020",
-    "allowJs": true,
-    "moduleResolution": "node",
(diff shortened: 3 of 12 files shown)


test-app/tests/unit/-private/properties/create-test.ts
```diff
@@ -94,7 +94,7 @@ module('create', function () {
       // @ts-expect-error violate types to check if it fails in weakly-typed envs
       create('');
       assert.true(false, 'should error');
-    } catch (e) {
+    } catch (e: any) {
```
ro0gr, line 97: can we have `Error` instead of `any`?

test-app/package.json
```diff
@@ -36,6 +36,7 @@
     "@embroider/test-setup": "^1.8.0",
     "@glimmer/component": "^1.1.2",
     "@glimmer/tracking": "^1.1.2",
+    "@tsconfig/ember": "^2.0.0",
```
ro0gr, line 39: looks like we should remove this from the package.json or from the ember-try scenario. I tend to remove it from the `package.json` in scope of the current `test-app`

ro0gr (commented):
Overall looks great!

I'd just like to keep testing against the pre-existing `moduleResolution: "node"` + add the new `moduleResolution` test which is broken now.

test-app/tsconfig.json
```diff
@@ -1,47 +1,18 @@
 {
+  "extends": "@tsconfig/ember/tsconfig.json",
   "compilerOptions": {
-    "target": "es2020",
-    "allowJs": true,
-    "moduleResolution": "node",
```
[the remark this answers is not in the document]
mayatron: The `@tsconfig/ember` 2.0.0 version has `moduleResolution` set to `node` (see [here](https://www.npmjs.com/package/@tsconfig/ember/v/2.0.0)) so there's not really a change in that respect. It only changes the value to `bundler` with the 3.0.0 version ([here](https://www.npmjs.com/package/@tsconfig/ember/v/3.0.0)).

Thoughts?

test-app/package.json
mayatron: See comment above.

test-app/tests/unit/-private/properties/create-test.ts
mayatron: We can't use the type `Error` here, due to:

`Catch clause variable type annotation must be 'any' or 'unknown' if specified.ts(1196)`

This is an annoying issue with TypeScript, see a [long discussion thread here](https://github.com/microsoft/TypeScript/issues/20024). Alternatively, and probably more correct, we can type this as `unknown`.

ro0gr:
@mayatron I do like your change. I'm going to merge and release it tomorrow.

ro0gr (approved):

ro0gr:
Just published it as 2.1.0. Thanks for your effort!

PR #630, opened 2023-10-31
fix: proper type declaration for `count` attribute

littleredridingfox:
Hello!

I've found this small issue with return type declarations for `count` attribute while rewriting some of my tests to TS.

Link to the attribute source code for reference: https://github.com/san650/ember-cli-page-object/blob/master/addon/src/properties/count.js#L91


ro0gr (approved):
Thanks

ro0gr:
published as 2.1.3

littleredridingfox:
> published as 2.1.3

Awesome, thank you!

Issue #619, opened 2023-09-30
get rid of `$.param(`


ro0gr:
This is a part of #617. We should pick the necessary bits from the original [`$.param(`](https://github.com/jquery/jquery/blob/ace646f6e83e653f666ba715c200739f1cbdba52/src/serialize.js#L55-L92) implementation.

Probably related: #211

PR #632, opened 2024-01-08
#619 chore: get rid of `$.param`

pomm0:
fixes https://github.com/san650/ember-cli-page-object/issues/619.

In order to remain backwards-compatible I implemented the Array and Object handler similar as $.param is doing. Could also only use 

`(new URLSearchParams(queryParams)).toString()`

but that is building array and object differently:
Selection_463

```
const params = {
  user_id: 5,
  comment_id: 1,
  hello: 'world',
  lorem: 'ipsum',
  reply_ids: [1, 2],
  search: {
    author: 'ember',
    topic: 'lts',
    deep: {
      my_array: [99, 77],
      my_value: true
    }
  }
};
const URLSearch = new URLSearchParams(params)
const jquery = $.param(params)
decodeURIComponent(URLSearch)
decodeURIComponent(jquery)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,7 +23,7 @@ Inside any of the packages you can run:
 
 - `cd addon && pnpm start` – Builds the addon in "watch mode" so changes picked up by test app.
 - `pnpm test` – Runs the test suite on the current Ember version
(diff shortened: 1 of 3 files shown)


ro0gr:
Hey, this looks great! I'll take a closer look on it this weekend. Thank you.

PR #636, opened 2024-01-29
Add typesVersions for resolving all type imports using legacy moduleResolution

simonihmig:
When TS is set up to not use a `moduleResolution` of `node16` or `bundler` yet (like legacy `node`/`node10`), then TS would not look into `exports`.

There is `types`, but that only works when resolving the main index module, like `import from 'ember-cli-page-object'`, but all other submodules like `ember-cli-page-object/adapter` would fail to resolve to `types.d.ts`. 

This is removing `types` in favor of `typesVersions`, supporting the same wildcard-matching that `exports` does.

### DIFF
--- a/addon/package.json
+++ b/addon/package.json
@@ -41,7 +41,11 @@
       "types": "./types.d.ts"
     }
   },
-  "types": "types.d.ts",
+  "typesVersions": {
+    "*": {
+      "*": ["./types.d.ts"]
+    }
+  },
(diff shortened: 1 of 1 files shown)


ro0gr:
Great! Thank you!

ro0gr:
published https://github.com/san650/ember-cli-page-object/releases/tag/v2.2.1

PR #638, opened 2024-02-02
Support string getters within collections

### LINKED ISSUE #637: string getters on a collection definition raise error
We have a large number of string props (selector strings) on page objects. While there are certainly better ways to use POs and write tests, _without_ those string props, doing the upgrade to v2 _and_ removing that tech debt at the same time is impossible, due to the scale of our code base. 

So for now I am applying the escape hatch described in the [deprecation guide](https://ember-cli-page-object.js.org/docs/v1.17.x/deprecations#string-properties-on-definition) and turning those into getters (codemod). So far so good. But in one case this approach fails, where you have a definition with a getter as part of a `collection()`. When accessing a single item of that collection via `objectAt`, that definition is cloned [here](https://github.com/san650/ember-cli-page-object/blob/28b1d8b29cf038927b5340e0c3f54b66ef358933/addon/src/properties/collection.js#L202), which "flattens" all getters into simple props. Later on, this simple string prop is then throwing that `string values are not supported in page object definitions` error.

As there is explicit support for getters, I think we should be able to make a clone of a definition that retains all own getters ([example](https://stackoverflow.com/a/34481052))!? @ro0gr does that make sense?

simonihmig:
Preserve getters within collections by applying their descriptors when cloning. 
Fixes #637

### DIFF
--- a/addon/src/-private/helpers.js
+++ b/addon/src/-private/helpers.js
@@ -209,3 +209,13 @@ export function findClosestValue(node, property) {
     return findClosestValue(parent, property);
   }
 }
+
+export function assignDescriptors(target, source) {
+  Object.getOwnPropertyNames(source).forEach((key) => {
+    const descriptor = Object.getOwnPropertyDescriptor(source, key);
+
+    Object.defineProperty(target, key, descriptor);
+  });
+
+  return target;
+}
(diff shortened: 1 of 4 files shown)


ro0gr:
Great! Thank you!

ro0gr:
published as v2.2.2

PR #639, opened 2024-02-16
Support plain arrays

simonihmig:
When having plain arrays defined on a PO, the string properties assertion would throw an error:
```
Error: string values are not supported in page object definitions Key: "0"`
```


ro0gr:
Thanks for the fix! I'm sorry you face issues like that. I'm going to merge and publish it by Friday.

ro0gr:
just published it under [v2.2.3](https://github.com/san650/ember-cli-page-object/releases/tag/v2.2.3).

Issue #641, opened 2024-07-10
Unknown blueprint: page-object | page-object-component


michaelchadwick:
Installed latest 2.30 version in an Ember 5.9 application, and tried running `ember generate page-object about-page`. I get the error: `Unknown blueprint: page-object`. Same type of issue with `ember generate page-object-component contact-form`.

ro0gr:
Hey, it should be fixed in the https://github.com/san650/ember-cli-page-object/releases/tag/v2.3.1 now. Thanks everyone!

PR #642, opened 2024-09-28
fix: include blueprints when publishing

xomaczar:
Closes https://github.com/san650/ember-cli-page-object/issues/641

### DIFF
--- a/addon/package.json
+++ b/addon/package.json
@@ -50,6 +50,7 @@
   },
   "files": [
     "addon-main.js",
+    "blueprints",
     "dist",
     "types.d.ts"
   ],


ro0gr:
Thanks for this! Unfortunately I could not find a way to extract the node tests in their own test app, to check this regression reliably, so I'm going to merge this change untested.

However CI here is currently failing for all the steps due to the pnpm/action-setup issues(https://github.com/san650/ember-cli-page-object/pull/643). Would you mind to rebase or pull from the master before merging this?

xomaczar:
Rebased

PR #643, opened 2024-09-29
fix(ci) upgrade pnpm/action-setup to the current latest

ro0gr:
it currently fails with the 
```
Running self-installer...
   WARN  GET https://registry.npmjs.org/pnpm error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
   WARN  GET https://registry.npmjs.org/pnpm error (ERR_INVALID_THIS). Will retry in 1 minute. 1 retries left.
   ERR_PNPM_META_FETCH_FAIL  GET https://registry.npmjs.org/pnpm: Value of "this" must be of type URLSearchParams
Error: Something went wrong, self-installer exits with code 1
```
https://github.com/san650/ember-cli-page-object/actions/runs/11084629777/job/30825358176?pr=642#step:4:13

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
         with:
           node-version: 14.x
 
-      - uses: pnpm/action-setup@v2.2.2
+      - uses: pnpm/action-setup@v2.4.1
         with:
           version: 7.29.3
           run_install: false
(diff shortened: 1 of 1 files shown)


ro0gr:
The original issue seems to be fixed, but it now fails for the release, beta, and canary scenarios with 
```
Object.hasOwn is not a function
```
https://github.com/san650/ember-cli-page-object/actions/runs/11096086617/job/30825526505?pr=643#step:6:62

I suspect it's not related to the ec-page-object, or at least I can't see an obvious "why", so let's go ahead and make the most of scenarios run and green.

ro0gr:
oh.. seems we need to upgrade to the node@16 to support the `Object.hasOwn(`, let's do it as a follow up then

PR #646, opened 2025-09-03
Fix package.json exports for index modules

simonihmig:
An import like `import { alias } from 'ember-cli-page-object/macros';` was failing to build when using Vite. It turns out there are indeed exports entries missing for all imports that should resolve to an index.js module. The existing `./*` would not cover those correctly (there is no `./dist/macros.js`).

I _think_ this change here is correct, I wonder though why this didn't came up earlier, with ember-cli and ember-auto-import or Embroider, both using webpack which should also acknowledge package.json exports. 🤔

### DIFF
--- a/addon/package.json
+++ b/addon/package.json
@@ -30,6 +30,21 @@
       "require": "./dist/index.js",
       "types": "./types.d.ts"
     },
+    "./adapters": {
+      "import": "./dist/adapters/index.js",
+      "require": "./dist/adapters/index.js",
+      "types": "./types.d.ts"
+    },
+    "./extend": {
+      "import": "./dist/extend/index.js",
+      "require": "./dist/extend/index.js",
+      "types": "./types.d.ts"
+    },
+    "./macros": {
+      "import": "./dist/macros/index.js",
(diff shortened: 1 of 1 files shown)


ro0gr:
Hey 👋 thanks for the PR.

I'm wondering if it makes sense to add vite mode to the test app. Do you know if is there a feasible way to add vite via ember-try? I'm mostly asking out of curiosity, your PR LGTM in its current state.

simonihmig:
> Do you know if is there a feasible way to add vite via ember-try

That would be nice, but I don't think that is practically feasible. There are too many differences in the boilerplate, it is not just changing dependencies. You could have a separate `test-app-vite` though, and run the same tests in both.

The latest [v2 addon blueprint](https://github.com/ember-cli/ember-addon-blueprint/tree/main) has a different setup with no separate test-app anymore, but it will _only_ run with Vite, and won't support classic or Emrboider/webpack.

ro0gr:
Just published it under v2.3.2 https://github.com/san650/ember-cli-page-object/releases/tag/v2.3.2

